PR #9, opened 2026-04-06
feat(map-projects): enrich custom match algorithm form

### LINKED ISSUE #2398: Validate Match API by calling GET to auto-fill algorithm metadata (id/name/description)
## User story
As a user, I want the Match API configuration to validate the endpoint by performing a GET request and auto-filling the algorithm metadata so configuration is easier and less error-prone.

## Use case
Configure a Match API by providing API URL and token, then confirm the endpoint is reachable and populate fields automatically.

## Requirements
- [ ] After the user provides API URL and token, perform a GET request to the Match API endpoint.
- [ ] The GET response must return: id (or ID), name, description.
- [ ] Auto-fill the UI fields (id/name/description) from the GET response.
- [ ] Keep POST on the same route for returning ranked candidates (current behavior).
- [ ] Require Match API implementations to support GET on the same route used for POST. (@filiperochalopes can update it on docs)

## Acceptance criteria
- [ ] Given a valid API URL and token, when the user saves/tests configuration, then the system performs GET and auto-fills id/name/description from the response.
- [ ] Given GET fails (e.g., auth/network/invalid endpoint), then the user sees a clear error and fields are not auto-filled.
- [ ] POST continues to work on the same route for candidate retrieval.

Ref: https://openmrs.slack.com/archives/C0A99PV25NG/p1771856424673959

filiperochalopes:
## Linked Issue

Closes https://github.com/OpenConceptLab/ocl_issues/issues/2398

### DIFF
--- a/src/components/map-projects/MultiAlgoSelector.jsx
+++ b/src/components/map-projects/MultiAlgoSelector.jsx
@@ -34,6 +34,7 @@ import orderBy from 'lodash/orderBy'
 
 
 import ConceptIcon from '../concepts/ConceptIcon'
+import APIService from '../../services/APIService';
 
 /**
  * MultiAlgoSelector (MUI5)
@@ -78,6 +79,7 @@ export default function MultiAlgoSelector({
   const { t } = useTranslation()
   const [expanded, setExpanded] = useState(() => new Map());
   const [errors, setErrors] = React.useState({})
+  const [customAlgoMeta, setCustomAlgoMeta] = React.useState({})
 
   const normalizedValue = useMemo(() => {
     let changed = false;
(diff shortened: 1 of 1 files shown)


src/components/map-projects/MultiAlgoSelector.jsx
```diff
@@ -219,6 +222,68 @@ export default function MultiAlgoSelector({
...
+    if(!selected?.url)
+      return;
+
+    const token = tokenOverride ?? selected?.token;
+    setCustomAlgoState(key, {isLoading: true, requestError: '', requestSuccess: '', tokenRequired: false});
+
+    const service = APIService.new();
+    service.URL = selected.url;
+
+    const response = await service.get(token || null, {}, undefined, true);
```
snyaggarwal: `token || false`

PR #12, opened 2026-04-21
OpenConceptLab/ocl_issues#2483 | Throttling Error UI

### LINKED ISSUE #2483: Mapper handling API throttling
## Summary

Handle API `429 Too Many Requests` generically across the app and show a throttling overlay with a retry countdown, localized messaging, and automatic recovery back to the same screen.

## Problem

Throttling can happen from many component-level API calls, not just app bootstrap. The frontend needs a generic handling path for `429` responses so users do not see crashes, broken views, or ambiguous messaging.

During testing, these issues came up:

- Initial handling only covered the `toggles()` request on app load.
- Component requests could still receive the `429` in normal success handlers and crash if they assumed a successful payload shape.
- The throttling message sometimes showed a generic message instead of saying whether the minute or daily limit was exhausted.
- The cooldown flow should preserve the user’s place instead of replacing the current page with a separate error route.
- Some throttling messages were hardcoded instead of coming from translations.

## Expected Behavior

- Detect `429` globally for normal API requests.
- Read `Retry-After` and show a live countdown.
- Read `X-LimitRemaining-Minute` and `X-LimitRemaining-Day` and show whether the exhausted limit is:
  - minute
  - day
  - both
- Show throttling as an overlay so the current page stays mounted underneath.
- Keep the overlay semi-transparent so the user can still recognize the underlying screen.
- When the countdown ends, automatically dismiss the overlay and let the user continue where they left off.
- Show the “rate limit ended” message via i18n, not hardcoded text.
- Keep throttling strings localized in `en`, `es`, and `zh`.

## Implementation Notes

- `APIService` now handles `429` generically and notifies the app through a shared throttle listener.
- `App` subscribes once to the throttle event and renders a throttling overlay.
- The overlay uses `Retry-After` for the ticking countdown.
- The overlay does not unmount the active route/component tree.
- On expiry, the overlay dismisses automatically and shows a localized success alert.
- `raw=true` request behavior remains intact for callers that intentionally inspect raw error responses.
- Defensive guards were added so pages do not crash if a throttled or unexpected payload shape slips into page logic.

## Files Touched

- `src/services/APIService.js`
- `src/components/app/App.jsx`
- `src/components/errors/ThrottlingError.jsx`
- `src/components/map-projects/MapProjects.jsx`
- `src/i18n/locales/en/translations.json`
- `src/i18n/locales/es/translations.json`
- `src/i18n/locales/zh/translations.json`

## API Dependency / Known Limitation

If the frontend still cannot determine whether the minute or day limit was exhausted, the browser is likely not receiving those headers in readable form.

The API response needs to expose:

- `Retry-After`
- `X-LimitRemaining-Minute`
- `X-LimitRemaining-Day`

If requests are cross-origin, the API must include these in `Access-Control-Expose-Headers`.

## Acceptance Criteria

- Any API call that returns `429` shows the throttling overlay.
- The countdown matches the `Retry-After` value.
- The overlay message correctly says whether the minute, day, or both limits were exhausted.
- The current page remains mounted beneath the overlay.
- The overlay is translucent enough that the underlying page is visible.
- When the timer ends, the overlay disappears automatically.
- A localized “rate limit ended” message is shown.
- No component crashes when a `429` happens mid-flow.
- `en`, `es`, and `zh` translations exist for the throttling flow.

## Reproduction

1. Open any screen that makes API requests.
2. Force the API to return `429`.
3. Confirm the overlay appears instead of the page crashing.
4. Verify the countdown uses `Retry-After`.
5. Verify the message reflects minute/day exhaustion from the limit headers.
6. Wait for the timer to reach zero.
7. Confirm the overlay disappears and the user can continue on the same screen.

Screenshot:

Image

Image

snyaggarwal:
## Linked Issue

Closes OpenConceptLab/ocl_issues#2483

Screenshots are in ticket

### DIFF
--- a/src/components/app/App.jsx
+++ b/src/components/app/App.jsx
@@ -17,7 +17,7 @@ import Footer from './Footer';
 import DocumentTitle from "./DocumentTitle"
 import './App.scss';
 import { hotjar } from 'react-hotjar';
-import APIService, { getThrottlingDetails } from '../../services/APIService'
+import APIService from '../../services/APIService'
 import Header from './Header';
 import OIDLoginCallback from '../users/OIDLoginCallback';
 import { OperationsContext } from './LayoutContext';
@@ -101,7 +101,7 @@ const App = props => {
 
   React.useEffect(() => {
     const unsubscribe = APIService.onThrottle(details => {
-      setThrottlingError(getThrottlingDetails(details.response || details))
+      setThrottlingError(details)
     })
 
     forceLoginUser()
--- a/src/components/app/App.jsx
+++ b/src/components/app/App.jsx
@@ -10,19 +10,21 @@ import Error403 from '../errors/Error403';
 import Error401 from '../errors/Error401';
 import WaitListing from '../errors/WaitListing'
 import NetworkError from '../errors/NetworkError'
+import ThrottlingError from '../errors/ThrottlingError'
 import ErrorBoundary from '../errors/ErrorBoundary';
 import CheckAuth from './CheckAuth'
 import Footer from './Footer';
 import DocumentTitle from "./DocumentTitle"
 import './App.scss';
 import { hotjar } from 'react-hotjar';
-import APIService from '../../services/APIService'
+import APIService, { getThrottlingDetails } from '../../services/APIService'
 import Header from './Header';
 import OIDLoginCallback from '../users/OIDLoginCallback';
 import { OperationsContext } from './LayoutContext';
 import Alert from '../common/Alert';
 import MapProject from '../map-projects/MapProject'
 import MapProjects from '../map-projects/MapProjects'
+import { useTranslation } from 'react-i18next';
 
 const AuthenticationRequiredRoute = ({component: Component, ...rest}) => {
   const { toggles } = React.useContext(OperationsContext);
@@ -47,7 +49,9 @@ const AuthenticationRequiredRoute = ({component: Component, ...rest}) => {
 }
 
 const App = props => {
-  const [networkError, setNetworkError] = React.useState(false)
+  const { t } = useTranslation();
+  const [startupError, setStartupError] = React.useState(null)
+  const [throttlingError, setThrottlingError] = React.useState(null)
   const { alert, setAlert, setToggles } = React.useContext(OperationsContext);
   const setupHotJar = () => {
     /*eslint no-undef: 0*/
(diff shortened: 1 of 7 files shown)


paynejd (commented):
## Review — two bugs to fix before merge

### 1. Translation interpolation mismatch (es/zh)

The English `limit_hit` strings are plain text, but the Spanish and Chinese translations include `{{minuteRemaining}}` and `{{dayRemaining}}` interpolation placeholders that are never passed from the component. `getLimitMessage` in `ThrottlingError.jsx` calls `t('errors.throttled.limit_hit.minute')` without any interpolation values.

As-is, Spanish and Chinese users will see literal `{{minuteRemaining}}` in the UI.

**Fix:** either remove the `{{...}}` placeholders from `es` and `zh` to match `en`, or pass the remaining values through to the `t()` call (which would require `ThrottlingError` to receive `minuteRemaining`/`dayRemaining` as props).

### 2. Double-processing of throttling details in App.jsx

In `App.jsx`, the `onThrottle` callback does:
```js
const unsubscribe = APIService.onThrottle(details => {
  setThrottlingError(getThrottlingDetails(details.response || details))
})
```

But `notifyThrottlingListeners` already calls `getThrottlingDetails` internally and passes the processed result to the listener. So `details` is already a throttling-details object (with `retryAfter`, `limitType`, etc.) — it has no `.headers` property. Wrapping it in `getThrottlingDetails` again extracts headers from a plain object that doesn't have any, so everything falls back to defaults (`retryAfter: 0`, `limitType: 'unknown'`).

**Fix:** use the details directly:
```js
const unsubscribe = APIService.onThrottle(details => {
  setThrottlingError(details)
})
```

---

Everything else looks good — architecture is clean, overlay-not-route is the right UX call, and the i18n coverage is solid. These two should be quick fixes.

paynejd (approved):
Both issues from my previous review are addressed:

1. **Translation interpolation mismatch (es/zh)** — fixed; all three locales now use plain text without `{{...}}` placeholders.
2. **Double-processing in App.jsx** — fixed; `setThrottlingError(details)` is now called directly so `retryAfter` / `limitType` flow through correctly.

Ticket acceptance criteria are all met. A few minor non-blocking observations for a possible follow-up:

- Dead `!== 429` branch in `fetchToggles` — 429s never reach the `.then` handler since `APIService` converts them to a never-resolving promise.
- `ThrottlingError` countdown effect recreates its interval every tick (depends on `secondsLeft`); pinning to `[retryAfter]` would be cleaner.
- `createPendingRequest()` means `.finally()` cleanup in callers won't fire on 429 — fine in practice since the overlay covers the page.
- Overlay has no `role="dialog"` / `aria-live` — minor a11y polish.

LGTM.

PR #15, opened 2026-05-05
OpenConceptLab/ocl_issues#2383 | AI Assistant Drop Down for Prompt + model

### LINKED ISSUE #2383: [OCL Mapper] Power User AI Assistant Drop-Down for Model & Prompt Template Selection
## Summary
Add a configuration drop-down to the OCL Mapper AI Assistant, visible only to users with the `core_user` role, that allows selection of:
1. **AI Model** - similar to current behavior, however the default selection may be different for each prompt
2. **Prompt Template** - populated by calling the new AI Assistant `/prompts/` endpoint with filter `action_type=match_recommend`; default value for this is where the `prompt_key` matches the `action_type`, so it's the same as the current default `match_recommend`

## Default / Fallback Behavior
Users without the required role are unaffected. The AI Assistant continues to use prompt template key `match_recommend` with action type `match_recommend` as it does today. This is also the default selection for eligible users who haven't made a choice.

## Notes
- Prompt templates sharing the same `action_type` are expected to share the same input/output schema, so templates returned for `match_recommend` are schema-compatible and safe to swap in.

snyaggarwal:
## Linked Issue

Ref OpenConceptLab/ocl_issues#2383


paynejd (changes requested):
## Prompt Template Persistence & Version Resolution

The UI selection work looks good, but there's a significant gap: **the prompt template + model selection is not persisted on the project**. Currently it's ephemeral React state that resets on reload. The intended behavior:

### 1. Project config should persist `prompt_template_key` + default model

The `MapProject` needs to store which prompt template key (e.g. `match-recommend`) and model the user selected as defaults for that project. This likely requires an API-side field on `MapProject` in oclapi2. On page load, the project's saved key should be the initial selection — not just falling back to `PROMPTS_KEY_DEFAULT` every time. The global default for this is -- including for map projects that do not have their project default defined -- is match-recommend (no prompt template version is resolved as latest at runtime) using the default model for the match-recommend template. This global default value does not actually need to be set for each project, since the Mapper should be able to handle this runtime. For sysadmins, the ability to change the default template might be useful, so that should be a global config value.

### 2. Version resolution should happen at invocation time, not page load

Currently the version comes from whatever `/prompts/` returned when the page loaded. Instead:

- **Config stores only the key** (no version) — so when an OCL admin publishes an improved prompt template, users automatically get it without changing their config.
- **At invocation time** (when AI assistant is called, or at the **start** of an auto-match run), resolve the current version from the API. This ensures a single auto-match run uses a consistent prompt template version throughout, even if the template is updated on the backend mid-run.
- **Log the fully-resolved `prompt_template_uri`** (e.g. `/prompts/match-recommend/v3/`) with each AI response for full auditability.

### 3. Summary of expected data flow

| Layer | What's stored | When |
|-------|--------------|------|
| Project config | `prompt_template_key` + `model_id` | User saves config; Optional with fallback to global default |
| Runtime (per run) | Resolved `prompt_template_uri` (key + version) | Start of auto-match |
| AI response log | Full `prompt_template_uri` + model | Each AI response |

Version pinning (letting users pick a specific version) is not MVP — future feature.

---

## Dead code: `AIPromptTemplateSelector.jsx`

`AIPromptTemplateSelector.jsx` is added in this PR but never imported or used anywhere. Should be removed, or if it's planned for a future integration point, defer it to that ticket.

paynejd (changes requested):
## Blocker: committed OIDC client secret in `webpack.config.js`

```js
'process.env.LOGIN_REDIRECT_URL': JSON.stringify('http://localhost:4004'),
'process.env.OIDC_RP_CLIENT_ID': JSON.stringify('oclmaplocal'),
'process.env.OIDC_RP_CLIENT_SECRET': JSON.stringify('dnLXaMWK3XdXkuvbXB6ABq7mSdGbuUWL'),
```

These replace `env.*` references with hardcoded local dev values, including a literal OIDC client secret. Even if `oclmaplocal` is only used against a local Keycloak, secrets should not be in source control, and the URL/ID belong in env config. Prior commit `d30096a` ("reverted env vars") shows this has happened before — please revert these three lines before merge.

Once that's reverted, the rest of the PR is good to go — prompt template persistence, invocation-time version resolution, and resolved URI logging all look right.

PR #16, opened 2026-05-06
Unified mapper data model PR1: normalizers + parallel state

paynejd:
## Summary

PR1 of the row-scoped, canonical-URL-identified candidate/concept refactor for OCL Mapper. Introduces the new entity model (AlgorithmResponse, Candidate, ConceptDefinition, ConceptRow) and the normalizer that produces it, behind a feature flag that is **OFF by default**. Reads still come from legacy `allCandidates` — flipping reads is PR2.

Spec: [unified-mapper-model.md](https://github.com/OpenConceptLab/ocl-online-docs/blob/main/mapper/unified-mapper-model.md) — full design rationale, key decisions, and implementation status table.

References OpenConceptLab/ocl_issues#2337 (does not yet close — PR1 of 3).

## What's in this PR

**New files:**
- `src/components/map-projects/normalizers.js` — pure `normalizeAlgoResult` / `normalizeAlgorithmInvocation`. Resolves canonical `ConceptReference {url, code, version?}` per algorithm's `concept_identity` config (`target_repo` / `bridge_repo` / `fixed`). Produces the four entities with intra-invocation dedup (richer ConceptDefinition wins).
- `src/components/map-projects/conceptKey.js` — opaque pool key helpers. Avoids the FHIR canonical `|version` collision that would arise from concatenating `url|code`.
- `src/components/map-projects/__tests__/normalizers.test.js` — 26 unit tests (Node `node:test`, no new framework dependency).

**Modified:**
- `src/components/map-projects/algorithms.jsx` — `concept_identity` block on `ocl-search` and `ocl-semantic`.
- `src/components/map-projects/MapProject.jsx` — `UNIFIED_MODEL_ENABLED` feature flag (default OFF), parallel `rowMatchState` hook, `mergeIntoRowMatchState` callback, `buildProjectContext` (reads `repo.canonical_url` or derives `https://ns.openconceptlab.org{url}`), flag-gated normalize-and-merge in `onResponse` (success + failure paths).
- `package.json` — `npm test` script using built-in `node --test`.

## Verification

- `npm test` → 26/26 passing
- `npm run eslint` → clean
- `npm run build` → webpack compiled successfully (after `npm install` to sync recent package updates)
- Smoke: with the flag off, runtime behavior is unchanged

## Test plan

- [ ] Code review the normalizer + tests against the spec linked above
- [ ] CI: `npm install`, `npm test`, `npm run eslint`, `npm run build` all green
[...]

PR #13, opened 2026-04-24
OpenConceptLab/ocl_issues#2485 | Copy project

snyaggarwal:
## Linked Issue

Ref OpenConceptLab/ocl_issues#2485

PR #8, opened 2026-04-06
OpenConceptLab/ocl_issues#2388 | custom encoder model for reranking

snyaggarwal:
## Linked Issue

Ref OpenConceptLab/ocl_issues#2388

PR #7, opened 2026-04-03
OpenConceptLab/ocl_issues#2315 | Updated Search Highlights dialog

snyaggarwal:
Reworks the Search Highlights dialog per the acceptance criteria in https://github.com/OpenConceptLab/ocl_issues/issues/2315:

- Renamed "Search Highlight" → "Match Details", "Search Score" → "Unified Score", "Search Raw Score" → "Raw Scores"
- Moved close button to top-right X (CloseIconButton)
- Scores displayed using color-coded chips consistent with the candidates tab, with per-algorithm raw score breakdown
- Added "Matched Attributes" section header; category names as sub-headers
- Improved synonyms readability (smaller font, tighter line-height, increased spacing)
- Extracted getScoreDetails and ScoreValueChip from Score.jsx for reuse
- Fixed truthy check on CIELMappedSources.length (was length, now length > 0)

PR #18, opened 2026-05-11
PR2b: read-side flip + ensureLoaded + UI config + flag ON (#2337)

paynejd:
## Summary

PR2b of the unified candidate/concept data model refactor for [OpenConceptLab/ocl_issues#2337](https://github.com/OpenConceptLab/ocl_issues/issues/2337). Flips reads from the legacy `allCandidates` shape to the new `rowMatchState + conceptCache` model via structured tuples, lands `$resolveReference`-backed `$lookup` as `ensureLoaded`, surfaces canonical-URL / namespace / bridge_repos[] in the project config UI, debounces the rerank trigger, and turns `UNIFIED_MODEL_ENABLED` on as the last step.

Builds on [#16 (PR1)](https://github.com/OpenConceptLab/oclmap/pull/16) and [#17 (PR2a)](https://github.com/OpenConceptLab/oclmap/pull/17). Spec: `plans/unified-mapper-model.md`.

---

## ⚠️ DO NOT MERGE TO `main` UNTIL STAGING VERIFICATION COMPLETES

This PR turns `UNIFIED_MODEL_ENABLED` to `true`. **Bridge**, **scispacy**, and **AI Assistant** code paths only attach at production build time via `PRIVATE_PACKAGES_GIT` and could not be exercised in local OSS dev (`BridgeMatchStub.canBridge() === false`). They must be verified on the staging deployment before this PR is merged to `main`.

**Required staging checks** (details in the test plan below):
1. Bridge cascade fan-out end-to-end (CIEL → LOINC)
2. Scispacy `$resolveReference` flow
3. AI Assistant recommendation with v2 payload

If any of these fail on staging, hold the merge and triage before promotion to prod.

Suggested workflow: merge this branch to a staging branch (or however the staging build is triggered), exercise the unchecked items below on the staging URL, and only then promote to `main`.

---

## What's in this PR

**Stage A — write-side foundations**
- Fix latent PR1 keying bug in `mergeIntoRowMatchState`: `def.url`→`def.key`, `cr.concept_url`→`cr.concept_key`. Was silent because the flag was OFF.
- Extract `normalizeLegacyAllCandidates` into `normalizers.js` and call it from `fetchAndSetProject` so reloaded v1 projects backfill `rowMatchState` under the flag. Necessary precondition for the flag flip; PR3's schema-v2 load subsumes this.

**Stage B — `ensureLoaded` over `$resolveReference`**
- `ensureLoaded(conceptKeys)`: batched `POST /$resolveReference/?namespace=...` followed by per-resolved concept fetch, with in-flight `Promise` dedup keyed by `concept_key`. Replaces `lookupCandidates` / `lookupCode`. Direct `ocl_url` fetch when available, else canonical resolution path.

**Stage C — UI config**
- `MultiAlgoSelector`: new required `canonical_url` field for custom algos with URL validation; bridge canonical_url surface; `lookup_required` checkbox removed (no longer a property of the algorithm).
- `ConfigurationForm`: target_repo canonical URL displayed with "Auto-derived" badge when `canonical_url_source === 'derived'`; bridge_repos[] summary with same badge convention; `Resolution Namespace` under an Advanced disclosure (defaulted to project owner).
- `namespace` state plumbed through `MapProject`; persisted on save/load.

**Stage D — read flip to structured tuples**
- New `viewBuilders.js` (pure JS, node:test loadable): `buildAlgorithmRowViews`, `buildQualityRowViews`, `candidateToRowView`, `sortRowViews`, `conceptForMapping`, `getScoreDetails`, `resolveAICandidateID`.
- `Candidates.jsx`: `getCandidates` rewritten to iterate `RowState.candidates` (algo view) / `RowState.concept_rows` (quality view); passes `{candidate, conceptDefinition, conceptRow, bridgeConceptDefinition?, bridgeChildren?, contributingCandidates?}` tuples downward.
- `Concept.jsx`: receives tuples directly; ad-hoc `mapping.cascade_target_*` extraction removed; bridge children nested under their parent bridge in algorithm view, surface as their target concept in quality view.
- `Score.jsx`: reads `conceptRow.rerank_score` + `candidate.score`; thin wrapper over the pure `getScoreDetails` adds the `bucketColor` UI affordance.
- `setAutoMatched` and `setStateViews` rewritten to read from `rowMatchState + conceptCache` via a `pickTopRowView` helper that filters by `target_repo.canonical_url` so bridge intermediaries are excluded from auto-mapping.
- Bulk match path (`processBatch.then`) now normalizes and merges into `rowMatchState` before `setStateViews` consumes it.

**Stage E — rerank debounce + in-flight check**
- `scheduleRerank(rowIndex)`: debounce timer per row + in-flight `Set` + rerun-needed flag for the case where new `ConceptRow`s arrive while an earlier rerank is still flying. Triggered from `mergeIntoRowMatchState` whenever any `ConceptRow` has `rerank_score === undefined`. Replaces the "wait for every algo" implicit batching.
- Rerank writes `rerank_score` directly onto `ConceptRow`; legacy `allCandidates` write preserved for save format compatibility (until PR3).
- Normalizer Stage E lift: `newConceptRow` now extracts `result.search_meta.search_normalized_score` onto `ConceptRow.rerank_score` so the single-algo `reranker:true` `$match` path doesn't need a separate `$rerank` round-trip.

**Stage F — AI Assistant response resolution**
- `resolveAICandidateID`: prefer `primary_candidate.concept_key` resolved via `conceptCache[concept_key].reference.code`; falls back to PR2a's `canonical_reference.code` shim, then legacy `concept_id` / `id`. Used in both Candidates.jsx's UI highlight match and MapProject's CSV export.

**Stage G — flag flip**
- `UNIFIED_MODEL_ENABLED = true` at MapProject.jsx.

### ✅ Local verification (done)

- [x] `npm test` — 79/79 (was 27, +52 new tests)
- [x] `npm run eslint` — clean
- [x] `NODE_ENV=production npm run build` — green
- [x] `docker compose build` — green (production target)
- [x] **Save/load round-trip** end-to-end against prod API from local dev:
  - 4 matched rows preserved across save → hard refresh → reload
  - `rerank_score` values byte-identical (e.g. `81.87800645828247` → `81.87800645828247`)
  - `allCandidates` envelopes restored with full `results` arrays
  - `rowMatchState` rehydrated with same candidate/concept_row counts
  - No `$match` re-fire on row open for previously-matched rows
- [x] Single-algo `ocl-search`: candidates render, scores show, mapping persists
- [x] Configuration form: canonical URL badge, namespace under Advanced, bridge canonical surfaces
- [x] Bridge fan-out at the **view layer** unit-tested (data shape exercised via fixtures; live algo waits for staging)

### 🟡 Staging verification — REQUIRED BEFORE MERGE TO `main`

These code paths only attach at production build time. **Each must pass on staging before this PR is promoted to prod.**

- [ ] **Bridge** (CIEL → LOINC) — per [PR2a's verification notes](https://github.com/OpenConceptLab/oclmap/pull/17):
  - `rowMatchState` populates with bridge intermediary + cascade targets at correct canonicals (CIEL: `https://CIELterminology.org`, LOINC: `http://loinc.org`)
  - AI payload `bridge_context` populates with `target_concept_keys` pointing at the LOINC entries in `recommendable_concepts`
  - Quality view shows the LOINC target with both contributing candidates while the CIEL bridge appears as its own bucketed concept
  - Selecting a `bridge_child` candidate maps to the LOINC target (not the CIEL intermediary)
  - Auto-Match never selects a bridge intermediary (the `pickTopRowView` filter by `target_repo.canonical_url` excludes them)
- [ ] **Scispacy** — `$resolveReference/?namespace=...` fires when scispacy returns LOINC codes; per-concept GET requests follow the resolved repo URL; concept cards fill in with names/descriptions (not just bare IDs)
- [ ] **AI Assistant** — request body shows v2 fields (`recommendable_concepts` + `bridge_context` + `target_repo`) alongside legacy `candidates`; `payload_version: 'v2'` present; `resolveAICandidateID` highlights the recommended candidate in the list with the sparkle icon

### Regression check (also on staging)

- [ ] Auto-match a project across all rows (multi-algo + bridge) → verify Mapped/Ready/Unmapped counts match expectations
- [ ] Open AI Assistant recommendation on a multi-algo + bridge row → verify primary_candidate is from the target repo (not a bridge intermediary)
- [ ] Save → reload an existing legacy v1 project → verify candidates render identically (the normalize-on-load path)

## New tests (52)

- `__tests__/views.test.js` (21) — algorithm grouping, quality grouping with bridge fan-out, multi-algo convergence, multi-bridge namespaces, sort orderings, missing-cache-entry handling.
- `__tests__/normalizeLegacyAllCandidates.test.js` (11) — load-path backfill including bridge cascade fan-out, multi-algo dedup, identity injection fallback, dedup across rows.
- `__tests__/viewHelpers.test.js` (20) — `getScoreDetails` bucketing for recommended/available/low_ranked, `conceptForMapping` legacy projection incl. bridge_child, `resolveAICandidateID` resolution chain.

## Not in this PR (PR3 scope)

- Drop legacy `allCandidates` state
- Schema-v2 save format + `normalizeLegacy.js` for v1 backward-compatible load
- Drop legacy `candidates` field + `payload_version` + `concept_id`/`id` shims from AI Assistant payload
- Pagination append path under the unified model

## Coordination

- The `ocl-ai-assistant` prompt template revision (separate session) reads the v2 payload fields and structurally excludes bridges from the recommendation pool. PR2b can ship without it — legacy `candidates` is still in the payload (PR2a Option A).
- The `_to_essential` allow-list addition in `ocl-ai-assistant` for `recommendable_concepts` + `bridge_context` ships alongside the prompt template revision.


paynejd:
## Review-feedback follow-up commit (d7180d7)

Addresses correctness gaps + UX polish found while testing the read-flip locally against a pre-PR2b legacy project. No new tests required — covered by the existing 79/79 suite. Bridge / scispacy / AI Assistant staging verification still gates merge per the original PR description.

### Correctness
- **#1 — `pickTopRowView` canonical filter**: now reads `target_repo.canonical_url` from `buildProjectContext()`, the same source the normalizer uses to stamp `ConceptDefinition.reference.url`. Eliminates the derived-vs-explicit mismatch that could silently empty auto-match.
- **#2 — Custom-algo `canonical_url`**: was decorative — user could enter a URL but nothing wired it into `concept_identity`. Centralized injection in new `ensureConceptIdentity()` helper (`algorithms.jsx`), used by `getAlgoDef`, bulk `processBatch.then`, `lookupCandidates`, and the legacy load path. Also adds Save-time validation: Save button disabled and an Alert lists offending algos when canonical_url is missing/malformed.
- **#7 — Rerank fail-loud**: `matchRerankResultToKey` throws on canonical-identity miss instead of silently falling back to fuzzy `ocl_url` / `id+source` matches. The rerank loop collects errors, logs a single summary entry, and surfaces an alert to the user.
- **#8 — State-driven `ensureLoaded`**: `mergeIntoRowMatchState` now auto-triggers `ensureLoaded` for newly-arrived ConceptDefinitions with `lookup_status !== 'full'` (bridge cascade targets, sparse algorithm results). Uses a forward-ref pattern mirroring `scheduleRerankRef`.
- **#13 — Missing target canonical on load**: when a saved project has match data but no resolvable target canonical, surface an error alert and force-open the configuration drawer instead of silently rendering an empty candidate list.

### Click-path wiring (rowView ↔ legacy concept shape at the seam)
- Candidate row click → details modal: `Candidates.jsx` decorates rowViews with top-level `id`/`url`/`version_url` so `SearchResults.handleRowClick` can resolve clicks back to the rowView and fire `onShowItemSelect`.
- Score chip click → Matched Attributes: `Concept.jsx` passes the `conceptForMapping` projection (not the raw tuple) to `setShowHighlights` so `SearchHighlightsDialog` finds `search_meta.search_highlight`.
- Score chip unified score regression: `viewBuilders.getScoreDetails` now accepts either the `{candidate, conceptRow}` tuple or the legacy `{search_meta}` shape so the dialog works regardless of which form the caller sends.

### CSV export
- `__map_repo_url__` and `__map_repo_id__` gated on `concept` (the `mapSelected` entry) so they're blank on unmapped rows — matches the rest of the `__map_*` namespace. Dropped the asymmetric project-repo fallback that predated PR2b.

### UI polish + i18n
- Compact canonical-URL caption under target repo: info icon + monospace URL with ellipsis truncation + tooltip + muted `· derived` suffix. Bridge repos use the same line per bridge. Replaces the orange warning-colored chip.
- All PR2b-introduced `t()` calls switched from `t(key) || 'default'` (which never falls through — i18next returns the truthy key string when missing) to `t(key, 'default')` / `t(key, {defaultValue, ...interp})`.
- Added the PR2b string set to `en/translations.json`. `resolution_namespace_description` now spells out the default behavior: *"When blank, defaults to {{owner}}"* — answers a reviewer question about whether namespace defaults to global or to the project owner (it's the project owner). Other locales inherit via `fallbackLng: 'en'`.

### Verified
- 79/79 unit tests pass, eslint clean, `NODE_ENV=production npm run build` green.
- Live in-browser: legacy project loads, candidates render, row clicks open details, score chips open the highlights dialog with unified score + matched attributes, Auto-Match maps a row whose `__map_algorithm__` correctly serializes to `ocl-semantic` via the unified-model projection.

snyaggarwal:
Tried against 2 rows with CIEL bridge algo for ICD-11-WHO

1. Score is not coming on top
Screenshot 2026-05-12 at 12 12 09
2. mapping to bridge target (ICD) is not getting logged
Screenshot 2026-05-12 at 12 10 45
3. Candidates table view not complete
Screenshot 2026-05-12 at 16 45 29
4. Target code on left is always empty
Screenshot 2026-05-12 at 12 12 09
5. Fetch More in Candidates is not doing anything (no calls in network, infinite loading candidates...)
Screenshot 2026-05-12 at 16 47 22

6. Auto Match not firing any calls with just bridge algo. With two algos (bridge + ES Token Search) it did.

7. Map a CIEL candidate from Unified candidates list and then go to algorithm view of candidates, no indicator in candidates that which CIEL concept was mapped

snyaggarwal:
Another thing:
This PR may send the rows to rerank API with no name (None) because bridge targets lookup may fail.
Reranker will assign -100000 score to those and they look weird in candidates list

snyaggarwal:
Trying 8 rows with only Semantic Algo, target CIEL -- Automatch worked here.

1. every candidate is duplicated:
Screenshot 2026-05-12 at 17 11 06
2. Fetch More button fires the call but doesnt update the candidates list

paynejd:
## Review-feedback follow-up commit (2d96c09) — @snyaggarwal

Thanks for the detailed report. Addressed 6 of 10 in this commit; 2 are expected to resolve transitively; 1 needs your diagnostics.

| # | Issue | Status |
|---|---|---|
| 9 | Every candidate duplicated | **Fixed** — mergeIntoRowMatchState now drops same-algo candidates on each invocation (mirrors the legacy `reject(...)` on allCandidates). Concept_rows pruned when no surviving candidate references them. |
| 5/10 | Fetch More re-fires + doesn't update | **Fixed** — pagination append branch feeds appended results into rowMatchState via a new `{append: true}` option so earlier pages stay put. |
| 4 | Target code on left always empty | **Fixed** — `Concept.jsx` now wraps legacy concept-shape input through a `legacyToRowView()` synthesizer at the top. One change covers Target Code column, decision tables, Search results, anywhere Concept is invoked with a legacy projection. |
| 7 | Mapped CIEL concept indicator missing in algorithm view | **Fixed** — bridge intermediary's algoScoreFirst row was hard-coded `isSelectedForMap={false}`. Now passes the real function so the Mapped indicator renders when the user mapped the intermediary from Unified view. |
| 8 | Rerank sent rows with no name → -100000 | **Fixed** — `buildRerankRowsForRow` filters out ConceptRows whose ConceptDefinition has no usable `display_name` (typically pending bridge cascade targets). scheduleRerank stays re-eligible so the row reranks once ensureLoaded completes. |
| 2 | Bridge target mapping not logged | **Fixed** — `_onMap` no longer gates the log entry on `concept?.url`. Bridge cascade targets can lack ocl_url until $resolveReference resolves them; log now fires on `url \|\| id` and surfaces `object_id` as fallback. |
| 1 | Score not on top (bridge case) | **Expected to resolve transitively** via #8: once bridge targets get rerank_scores from the (now non-poisoned) rerank pass, they sort to the correct position. Worth re-checking after this commit. |
| 3 | Candidates table view not complete | **Expected to resolve transitively** via #4: same root cause (Concept bailing on legacy shape). Worth re-checking. |
| 6 | Auto Match doesn't fire calls with bridge-only algo | **Pending — need diagnostics from you.** Code review doesn't show a smoking gun; `fetchBulkBridgeCandidates` looks correct and `canBridge` should be true in your env. When you re-test, could you share: (a) the browser console + network tab during the bridge-only Auto Match, (b) whether the row was `Unmapped` when you clicked it (the bulk path filters `rowStatuses.unmapped` when `autoMatchUnmappedOnly` is true — a stale `readyForReview` entry would skip the row silently), and (c) whether bridge-only with `autoMatchUnmappedOnly=false` works? |

### Notes on the duplicate fix (#9)

The merge now does **replace-by-algorithm**, not stack. If you had any reliance on the old behavior (e.g. multiple algorithms returning the same concept building up cumulative candidates), that still works — candidates from different algorithms are unaffected; only same-(rowIndex, algorithm) re-runs replace. Multi-algo convergence in Quality view continues to dedup by concept_key.

### Notes on the pagination fix (#10)

`mergeIntoRowMatchState` now takes an options arg: `mergeIntoRowMatchState(rowIndex, normalized, {append: true})`. The append path skips the same-algo drop and the orphan concept_row pruning. All existing callers default to append=false; only the `offset > 0` branch in `fetchAllCandidatesForRow.onResponse` uses append=true.

snyaggarwal:
11. (with bridge) on candidates visit, it makes rerank call every time. 
 - rerank -> save -> reload -> row click -> rerank again
 - rerank -> close -> open -> rerank
12. No candidates rendered even when api returned candidates (this is the row I never ran earlier to get candidates)
Screenshot 2026-05-12 at 18 00 59

13. AutoMatch ran and got the candidates but dint selected map when all criterias for top candidates matched

14. Score is still NAN
Screenshot 2026-05-12 at 18 04 31

paynejd:
## Round-2 review fixes (104a7c9) — @snyaggarwal

Addresses the 4 issues you reported on the previous commit plus 4 from local testing. Net result: Quality view shows only target-repo concepts (bridge intermediaries appear only in Algorithm view, per spec), no more NaN% chips, no rerank re-fire loop, table view renders.

| # | Issue | Status |
|---|---|---|
| **11** | Rerank fires on every row visit | **Fixed** — `scheduleRerank` now gates on whether the ConceptDefinition has a usable `display_name`. Rows that the previous-commit's `buildRerankRowsForRow` filter skips no longer stay eligible forever. `writeConceptCachePatch` chains into `scheduleRerank` when a lookup transitions to having a display_name, so once `ensureLoaded` completes for a pending bridge cascade target, the row gets a score. |
| **14** | Score shows `NaN%` | **Fixed** — `getScoreDetails` returns `rerankScore: ''` / `algoScore: ''` when no numeric value is present. The Target Code panel's chip will now render empty (or you can choose to hide it) rather than `NaN%`. |
| **13** | AutoMatch ran but didn't auto-select | **Expected to resolve transitively via #11** — cascade targets now receive rerank scores via the chained `scheduleRerank`, so `pickTopRowView`'s sort no longer parks them at `-1`. Worth re-testing. |
| **12** | First-time row click w/ bridge: API returned data, UI shows none | **Fixed** — `buildProjectContext` now falls back to `'/orgs/CIEL/sources/CIEL/'` when `bridgeAlgo.target_repo_url` is missing (matches the MultiAlgoSelector placeholder). Without this, the normalizer's `resolveReference` returned `null` for every bridge primary and `rowState` ended up empty. Auto-Match happened to work because of state-ordering quirks; manual click surfaced the underlying bug. |
| | Quality view should only show target-repo concepts | **Fixed** — `buildQualityRowViews` output is now filtered by `conceptDefinition.reference.url === target_canonical` (sourced from `buildProjectContext()`). Bridge intermediaries are gone from this view; they only appear in Algorithm view as parent groupings for their cascade children. |
| | Algorithm view shows Map button on every candidate | **Fixed** — Map action gated on the same `target_canonical` everywhere (action column + Concept.jsx Item). Bridge intermediary's algoScoreFirst row reverts to the placeholder spacer it had before; cascade children retain their Map action since they're in target_repo. |
| | Table view broken for all candidates | **Fixed** — `rowsForTable` now spreads the `conceptForMapping` legacy projection AND the rowView fields. Table cells read legacy fields (`names`, `descriptions`, `source`, etc.); card view renderer reads the tuple. Click-lookup still works (`id`/`url`/`version_url` set explicitly at the end). |

### Things still pending diagnostics

- **S6** (Auto Match no API calls with bridge-only algo) — your console/network log would help. If #12's `target_repo_url`-fallback fix happens to also resolve S6 (same bridgeAlgo state inspection), please confirm.

### Re-test request

When you have a minute, would you re-run your previous flows and confirm:
1. Bridge cascade in Algorithm view — bridge intermediary visible, no Map button on it, cascade children below with their Map buttons.
2. Quality view — only target-repo concepts (no CIEL intermediaries).
3. Open a row that hasn't been matched yet → candidates render (previously empty).
4. Open a previously-matched row → no extra rerank API call fires (previously fired every time).
5. Map a row, then re-open the Mapping Decisions panel — score chip shows a number (previously NaN%).
6. Switch to table view — full columns populated (previously blank).

paynejd:
## Round-3 commits pushed (fdc60b8…45416ea) — @snyaggarwal

16 targeted fixes from a full exercise of the read-flip path. Tests now 85/85, eslint clean, production webpack build green. Working tree clean.

### Display / scoring correctness

| Commit | Bug | What landed |
|---|---|---|
| [`fdc60b8`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/fdc60b8) | helper-text encoding + bridge canonical fetch | `i18next` `escapeValue: false` — Resolution Namespace helper no longer renders `&#x2F;`. MultiAlgoSelector now fetches bridge source repo metadata and populates `bridge_repo.canonical_url` from `response.data.canonical_url`. CIEL surfaces as `https://CIELterminology.org` instead of the derived `ns.openconceptlab.org/...`. |
| [`9fcdbb2`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/9fcdbb2) | 1 — interim 100% chip | `getScoreDetails`: dropped the `score * 100` fallback. When `rerank_score` is undefined, percentile stays undefined; Score.jsx renders a muted em-dash instead of a fake 100%. |
| [`f29a801`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/f29a801) | 4 — bridge_child missing unified chip in algo view | Dropped `noScore={algoScoreFirst}` for bridge_child. Score component naturally hides the raw chip when `candidate.score` is undefined. |
| [`832edc1`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/832edc1) | 5 — scispacy synthesized rerank score | `fromScispacyResultsToConcepts` no longer sets `search_normalized_score = composite_score * 100`. Real rerank now fills `rerank_score`. |
| [`8b41e72`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/8b41e72) | 7 — missing LOINC summary properties | `toConceptDefinition` now captures `property` (schema-specific dict) and `extras`. `inferLookupStatus` promotes to `'full'` when `property` is present OR `names` is populated (LOINC concepts often have no separate descriptions). Server already returned this with `verbose=true`; client was dropping it. |
| [`0f122e5`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/0f122e5) | 8 — interim 100% (deeper cause) | New `trustServerRerank` ctx flag through `normalizeAlgorithmInvocation → normalizeAlgoResult → newConceptRow`. Server emits `search_normalized_score` unconditionally; in multi-algo mode that's per-algo native score, not a unified rerank. Per-row $match: `!isMultiAlgo && algoDef.provider === 'ocl'`. Bulk processBatch: `!isMultiAlgo`. Legacy load: `true` (saved scores ARE rerank). Bulk bridge/scispacy default `false`. |

### Algorithm coordination / cache correctness

| Commit | Bug | What landed |
|---|---|---|
| [`b0e8d52`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/b0e8d52) | 2 — scispacy retrigger on 0 results | `fetchScispacyCandidates` gate uses `existingEntry !== undefined`, not `.length > 0`. Zero-result rows no longer re-fire on every tab visit. |
| [`6697c52`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/6697c52) | 11 — outer reuse check | `fetchAllCandidatesForRow` reuse-check uses `existingCandidates !== undefined`. Companion to Bug 2 — without this, scispacy-zero-results dispatched again and the "Running" indicator pinned forever. |
| [`ef22c18`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/ef22c18) | 10 — tab switch retriggers algos | `onDecisionTabChange`: fixed `firstAlgo.id` typo (was a no-op making cache lookup always miss); added in-flight guard checking `rowStageRef.current` for `status === 0`. Discuss → Candidates no longer re-fires. |
| [`0ae57dd`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/0ae57dd) | 9 — wasteful double-rerank | `buildRerankRowsForRow` skips ConceptRows that already have `rerank_score`. Scispacy-late arrivals get a focused second rerank batch. |
| [`eb5e3d6`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/eb5e3d6) | 15 — auto-match regression | Bug 9's filter made bulk `rerank(idx, true)` a no-op when debounced rerank had already scored every row → skipped `setAutoMatched`. Fix: when `rerankRows` is empty AND `isBulk`, still schedule `setAutoMatched`. |

### Bridge framing / UX

| Commit | Bug | What landed |
|---|---|---|
| [`571a12b`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/571a12b) | 3 — bridge framing in Quality view | Bridge-only target renders inline `CIEL:1234 [SAME-AS] LOINC:52767-1`. Multi-algo convergence renders standard with a `[i]` tooltip on the algo chip listing `via CIEL:1234 ... — SAME-AS`. `buildQualityRowViews` emits new `bridgeContributors` array. |
| [`4b019fd`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/4b019fd) | 6 — sort toolbar misbehavior | `onSort` simplified: dropped the `option === sortBy` clause that flipped same-option re-clicks to `rerank_score`. Bridge children now sort by the same key as their parent. |
| [`22eaa6a`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/22eaa6a) | 13+14 — chip + algo-view polish | Compact `MAP_TYPE_CHIP_SX` (18px height, 11px label, baseline alignment, 4/6px padding). Algo-view bridge_child no longer repeats the bridge prefix on every nested child (gate on `!algoScoreFirst`). First bridge child in algo view gets its top divider. |
| [`1cd8aee`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/1cd8aee) | 12 — Unranked Candidates bucket | Quality view bucketing checks score presence first: `undefined` → new `pendingRerank` bucket with gray `PENDING_RERANK_COLOR` chip. Rows auto-migrate to recommended/available/low-ranked once `rerank_score` lands. |

### Load-path correctness

| Commit | Bug | What landed |
|---|---|---|
| [`45416ea`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/45416ea) | 16 — saved project reload drops candidates | `fetchAndSetProject` calls `normalizeLegacyAllCandidates` synchronously with a fallback-derived canonical (`ns.openconceptlab.org/...`) before `fetchRepo` resolves with the real canonical (e.g. `http://loinc.org`). Quality view filter then matches nothing. Fix: `useEffect` re-runs the normalizer when the live canonical changes (`lastNormalizedCanonicalRef` guards idempotence). `onSave` now persists `target_repo.canonical_url` so future loads skip the round-trip. |

### Architectural takeaways (preserve these contracts going forward)

1. **`search_normalized_score` is two different things.** From `reranker=true` $match: real rerank score (trust). From any other path: per-algo native score (ignore). New `trustServerRerank` ctx flag in normalizers gates the difference.
2. **The Quality view filter is load-bearing.** Compares `view.conceptDefinition.reference.url === targetCanonical`. Any asymmetry between the canonical stamped on `reference.url` and the live canonical silently empties the view (Bug 16).
3. **`buildProjectContext` is the source of truth at runtime;** `loadProjectContext` is a fallback used during the sync load step before `fetchRepo` resolves. `lastNormalizedCanonicalRef` `useEffect` re-normalizes if they disagree.
4. **Rerank pipeline has two flavors with different side effects.** Debounced doesn't carry `setAutoMatched`; bulk does (Bug 15).
5. **Three independent reuse checks** must agree on "we already ran this algo for this row." Convention: entry presence (`!== undefined`), not result count. If a fourth path appears, follow the convention.

### New tests (85 total now)

- `normalizers.test.js`: Bug 7 (property capture, lookup_status='full' on verbose payload), Bug 8 (multi-algo ignores server normalized score; single-algo opt-in still works)
- `views.test.js`: Bug 3 (`bridgeContributors` populated on convergence, empty on bridge-only)
- `viewHelpers.test.js`: Bug 1 updated (interim → `percentile: undefined`, `rerankScore: ''`, raw `algoScore` still surfaces)
- `normalizeLegacyAllCandidates.test.js` untouched (legacy loader opts in to `trustServerRerank: true`)

### Re-test request

When you have a minute, would you re-run against this rebased branch? Particularly:
1. **S6** (your original "Auto Match no API calls with bridge-only algo") — Bug 12 (the `buildProjectContext` bridge `target_repo_url` fallback in `104a7c9`) may have incidentally fixed this; worth verifying.
2. **S1 / S3** (score not on top, candidates table view incomplete) — expected to resolve transitively via Bug 8 / Bug 7 / Bug 16; please confirm.
3. **Saved-project reload race** (Bug 16) — open a saved project against a target repo with a real canonical URL (e.g. LOINC → `http://loinc.org`); confirm Quality view populates.
4. **Bridge convergence framing** (Bug 3) — multi-algo + bridge against same target concept; confirm the algo chip carries the `[i]` tooltip with `via CIEL:... — SAME-AS`.
5. **Scispacy edge cases** (Bug 2 / Bug 5 / Bug 11) — rows with zero scispacy matches; rows where scispacy lands late.

paynejd:
## Rounds 4–7 summary + open items for review — @snyaggarwal

Wrap-up across the last four commits since the round-3 push. **86/86 tests, eslint clean, production webpack build green.**

### What landed since round 3

| Commit | Round | What |
|---|---|---|
| [`e2ccf71`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/e2ccf71) | 4 | Three follow-ups from gap review: always send `namespace` on save (cleared field propagates now); deterministic primary candidate in multi-algo convergence (sort by `candidate.score` desc, was first-hit on iteration order); `ensureLoaded` honors `lookupConfig.token` (was using session token only). |
| [`48dc6b5`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/48dc6b5) | 5 | AI POST 404 fix: single-row invocations no longer pin prompt-template version into the invoke URL (`/prompts/<key>/<version>/invoke/` → `/prompts/<key>/invoke/`). Bulk auto-match still pins, since consistency across rows matters there. Scispacy 503 handling: `isLoading` race resolved (was clearing before response arrived), error responses no longer cached as `results: []` (was poisoning the reuse-check). Refresh button now visible even on no-candidates state. |
| [`0e5805e`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/0e5805e) | 6 | Stripped heavy fields (`extras`, `names`, `descriptions`) from the legacy `candidates[]` field sent to the LLM. Updated scispacy alert copy. (Bridge_child indent attempt — reverted in next commit.) |
| [`f0681c4`](https://github.com/OpenConceptLab/oclmap/pull/18/commits/f0681c4) | 7 | Reverted bridge_child indent attempts (the layout I shipped wasn't what was wanted). |

### 🔴 Open items needing Sunny's eyes

1. **Bridge-as-first-algo doesn't fire** — @paynejd reports that when `ocl-ciel-bridge` is the first algorithm in a project's algo list, the per-row click doesn't fire the fetch. Reordering bridge to second position works. Cannot reproduce from code review — might be timing on `bridgeRef.current` initialization, or a state-ordering issue specific to your build. Console + network log from the broken state would unblock this. If you can repro, please paste the output of:
   ```js
   console.log({canBridge: window.__bridgeRef?.current?.canBridge?.(), algos: window.__algosSelected})
   ```

2. **Table view sparse for bridge candidates** — @paynejd reports that the table view of bridge candidates only shows id + display_name; other columns empty. My hypothesis is that the rendered rows are bridge intermediaries (CIEL, which lacks LOINC schema dict) — but I need a screenshot showing which view mode (Algorithm/Quality grouping) and which columns are empty before fixing.

3. **Bridge_child secondary text alignment** — reverted twice, both attempts wrong. Need a sketch or description of the desired layout (where COMPONENT/PROPERTY chips should align relative to the [SAME-AS] map-type chip and the LOINC code text).

### 🔴 Open item NOT in this repo

4. **AI Assistant `max_tokens` error** ("requested 64000, can only afford 62098") is a **server-side** config in the prompt template's max_tokens param — needs adjustment in `ocl-ai-assistant`, not here. The client-side cleanup (input metadata trimming, version-pin removal) is done.

### Status snapshot

- **25 commits** total on the branch
- All automated checks green
- Ready for merge consideration once items 1, 2, 3 above are resolved (or explicitly deferred to a follow-up PR), and after a final smoke pass against your prod build

snyaggarwal (approved):

PR #20, opened 2026-05-13
OpenConceptLab/ocl_issues#2505 | added config for use_lexical_variants

snyaggarwal:
## Linked Issue

Refs OpenConceptLab/ocl_issues#2505

### DIFF
--- a/src/components/map-projects/AdvancedSettings.jsx
+++ b/src/components/map-projects/AdvancedSettings.jsx
@@ -3,11 +3,13 @@ import { useTranslation } from 'react-i18next'
 import TextField from '@mui/material/TextField'
 import Collapse from '@mui/material/Collapse'
 import Button from '@mui/material/Button'
+import FormControlLabel from '@mui/material/FormControlLabel'
+import Checkbox from '@mui/material/Checkbox'
 import UpIcon from '@mui/icons-material/ArrowDropUp';
 import DownIcon from '@mui/icons-material/ArrowDropDown';
 
 
-const AdvancedSettings = ({ namespaceValue, setNamespace, defaultNamespace }) => {
+const AdvancedSettings = ({ namespaceValue, setNamespace, defaultNamespace, isCoreUser, useLexicalVariants, setUseLexicalVariants }) => {
   const { t } = useTranslation()
   const [open, setOpen] = React.useState(false)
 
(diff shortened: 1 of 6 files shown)


src/components/map-projects/AdvancedSettings.jsx
```diff
@@ -31,6 +33,20 @@ const AdvancedSettings = ({ namespaceValue, setNamespace, defaultNamespace }) =>
...
+            isCoreUser &&
+              <FormControlLabel
+                sx={{marginTop: '8px'}}
+                control={
+                  <Checkbox
+                    size='small'
+                    checked={Boolean(useLexicalVariants)}
+                    onChange={event => setUseLexicalVariants(event.target.checked)}
+                  />
+                }
```
paynejd: The label alone ("Use Lexical Variants") won't tell a core user what flipping this does. Even though the namespace TextField above has a `helperText` explaining $resolveReference behavior, this checkbox has nothing equivalent. Suggestion — wrap in a small FormControl + FormHelperText so the description shows below:

```suggestion
          {
            isCoreUser &&
              <div style={{marginTop: '8px'}}>
                <FormControlLabel
                  control={
                    <Checkbox
                      size='small'
                      checked={Boolean(useLexicalVariants)}
                      onChange={event => setUseLexicalVariants(event.target.checked)}
                    />
                  }
                  label={t('map_project.use_lexical_variants', 'Use Lexical Variants')}
                />
                <div style={{fontSize: '12px', color: 'rgba(0, 0, 0, 0.6)', marginLeft: '32px', marginTop: '-4px'}}>
                  {t('map_project.use_lexical_variants_description', 'Expand $match search to include English spelling variants (e.g. color/colour, leukemia/leukaemia) when matching concept names.')}
                </div>
              </div>
          }
```

(Inline `<div>` avoids adding a new MUI import; if you'd rather use `FormHelperText`, that works too — needs `import FormHelperText from '@mui/material/FormHelperText'`.)

paynejd (commented):
This is the companion to oclapi2#868 — closes the loop on `use_lexical_variants` end to end (project persistence → checkbox → $match request). Reviewed together they now form a complete change.

The wiring is clean and minimal. Two small notes:

**1. Cross-link missing in both PR bodies.** Neither this PR's description nor oclapi2#868 mentions the sibling. Worth a one-line `Companion: OpenConceptLab/oclapi2#868` in each so reviewers (and future archaeologists) see the pair.

**2. "Use Lexical Variants" is opaque without context.** Inline suggestion below — a short helper/description so even core users can tell what the checkbox does without digging through code.

Merge order doesn't matter: if this lands first, oclapi2 master ignores the unknown `variants` field; if oclapi2 lands first, oclmap master doesn't yet send it so default-disabled holds. No race.

PR #21, opened 2026-05-13
OpenConceptLab/ocl_online#89 | Support multiple AI Assistant outputs per input row

snyaggarwal:
## Linked Issue

Closes OpenConceptLab/ocl_online#89


paynejd (commented):
Reviewed with @paynejd. Four required changes before merge — inline suggestions cover the MapProject.jsx side; the display change is below.

1. **Capture `output_locale` on every entry** — it's part of the request payload and meaningfully affects the response, so it belongs on the saved record and in the attribute strip.
2. **Drop the URI line from the visual display** — once the resolved version is shown next to the template key, URI is redundant. Keep `prompt_template_uri` in the data model (audit log + JSON viewer still see it).
3. **Read the resolved prompt version back from the invoke response** (`response.data.template.version`). The intent is to record which version the AI Assistant *used* to generate the response, not which version (if any) we requested — clients can invoke with no version, let the assistant resolve, and read it back from the response.
4. **Drop store-side dedup entirely** — keep the full per-row history of AI responses. Move the "fire once" gate into the auto-match path only: auto-match fires once per row; a user can manually request as many additional AI recommendations as they want, even with the same (model, prompt, locale) combo.

---

**Display side — `AICandidatesAnalysis.jsx`** (outside the diff hunks, so written out here): replace the existing URI block (`analysis?.prompt_template_uri && <span>…URI:…</span>` around line 128) with an `output_locale` block:

```jsx
{
  analysis?.output_locale &&
    <span style={{marginRight: '4px', display: 'inline-flex'}}>
      <Typography gutterBottom sx={{ color: 'text.secondary', fontSize: 12, mb: 0 }} component='span'>
        {t('map_project.output_locale')}:
      </Typography>
      <Typography gutterBottom sx={{ color: 'text.primary', fontSize: 12, mb: 0 }} component='span'>
        {analysis.output_locale}
      </Typography>
    </span>
}
```

Add a `map_project.output_locale` translation key to en/es/zh alongside it.

---

**Backend verification:** please confirm the new array-shaped `analysis` field round-trips cleanly through save→reload end-to-end. The load path defensively wraps singletons into `[v]`, but I didn't verify the save path or the ocl-online DB column accept the array shape. A quick smoke test (create project → run AI analysis → save → reload → re-run on same row → confirm both entries persist and the pager shows 2/2) before merge would be reassuring.

src/components/map-projects/AICandidatesAnalysis.jsx
```diff
@@ -20,9 +22,21 @@ import compact from 'lodash/compact'
...
+  const [page, setPage] = React.useState(0)
+
+  const analysisArray = Array.isArray(analysisProp) ? analysisProp : (analysisProp ? [analysisProp] : [])
+  const total = analysisArray.length
+
+  // Jump to latest when a new entry is added
+  React.useEffect(() => {
+    if(total > 0)
+      setPage(total - 1)
+  }, [total])
```
paynejd: Track previous total so the pager only auto-jumps when a new entry is *appended*, not on every re-render where `total` re-evaluates.

```suggestion
  // Jump to latest only when total grows (a new entry was appended);
  // don't yank the user back when total holds steady on re-render.
  const prevTotalRef = React.useRef(0)
  React.useEffect(() => {
    if(total > prevTotalRef.current)
      setPage(total - 1)
    prevTotalRef.current = total
  }, [total])
```

src/components/map-projects/MapProject.jsx
```diff
@@ -3765,7 +3768,11 @@ const MapProject = () => {
         })
       )
     }
-    if(isNumber(__index) && repoVersion && !analysis[__index] && _candidates?.length > 0) {
+    const existingAnalyses = analysis[__index] || []
+    const selectedModelId = getSelectedAIModel()?.id || AIModel
+    const templateKey = promptTemplate?.key
+    const alreadyAnalyzed = existingAnalyses.some(a => a?.model === selectedModelId && a?.prompt_template?.key === templateKey)
+    if(isNumber(__index) && repoVersion && !alreadyAnalyzed && _candidates?.length > 0) {
```
paynejd, line 3852: Move "fire once per row" into the auto-match path only. Single-row user clicks should always invoke and append — the user can request as many additional recommendations as they want.

```suggestion
    // Auto-match (caller supplied resolvedPromptTemplate) fires once per row;
    // user-initiated single-row clicks always append a new entry to the
    // per-row analysis history.
    const isAutoMatch = Boolean(resolvedPromptTemplate)
    const existingAnalyses = analysis[__index] || []
    const alreadyAnalyzed = isAutoMatch && existingAnalyses.length > 0
    if(isNumber(__index) && repoVersion && !alreadyAnalyzed && _candidates?.length > 0) {
```

src/components/map-projects/MapProject.jsx
```diff
@@ -3860,7 +3867,12 @@ const MapProject = () => {
 
         markAlgo(__index, 'recommend', 1)
         log({created_at: timestamp, action: 'AIRecommendation', description: get(response.data, 'output.rationale') || get(response.data, 'rationale'), extras: {...response.data, model: selectedModel, prompt_template: promptTemplateRef, prompt_template_uri: promptTemplateRef?.uri}}, __index)
-        setAnalysis(prev => ({...prev, [__index]: {...response.data, model: selectedModel?.id || AIModel, model_name: selectedModel?.name, prompt_template: promptTemplateRef, prompt_template_uri: promptTemplateRef?.uri, timestamp: timestamp, user: user.username || user.id}}))
+        const newEntry = {...response.data, model: selectedModel?.id || AIModel, model_name: selectedModel?.name, prompt_template: promptTemplateRef, prompt_template_uri: promptTemplateRef?.uri, timestamp: timestamp, user: user.username || user.id}
+        setAnalysis(prev => {
+          const existing = prev[__index] || []
+          const filtered = existing.filter(a => !(a?.model === newEntry.model && a?.prompt_template?.key === promptTemplateRef?.key))
+          return {...prev, [__index]: [...filtered, newEntry]}
+        })
```
paynejd: Rolls in concerns 1, 3, and the store-side half of 4:
- **`output_locale`** captured on the entry.
- **Resolved version** read back from `response.data.template.version` — the ai-assistant invoke endpoint already returns `{template: {key, name, version}, output, metadata}` (see `ocl-ai-assistant/core/prompts/services.py`), we were just overwriting it with the nulled-out client-side ref.
- **URI** reconstructed as `/prompts/<key>/<resolved-version>/` once the version is known, so it reflects the resolved template rather than the bare `/prompts/<key>/`.
- **Always append** — no more filtering out prior entries with the same (model, template_key).

```suggestion
        const resolvedTemplate = response.data?.template || {}
        const resolvedVersion = resolvedTemplate.version || promptTemplateRef?.version || null
        const resolvedPromptRef = {
          ...promptTemplateRef,
          version: resolvedVersion,
          uri: resolvedVersion && promptTemplateRef?.key ? `/prompts/${promptTemplateRef.key}/${resolvedVersion}/` : (promptTemplateRef?.uri || null)
        }
        const newEntry = {...response.data, model: selectedModel?.id || AIModel, model_name: selectedModel?.name, prompt_template: resolvedPromptRef, prompt_template_uri: resolvedPromptRef.uri, output_locale: promptOutputLocale || null, timestamp: timestamp, user: user.username || user.id}
        setAnalysis(prev => ({...prev, [__index]: [...(prev[__index] || []), newEntry]}))
```

PR #22, opened 2026-05-14
PR3-X.2 | Drop inherited version slot on bridge cascade-target references

### LINKED ISSUE #2520: PR3-X.2 | Bridge cascade target inherits target_repo version 'HEAD', splitting concept_key from direct-path matches
## Summary

A target-repo concept that arrives via BOTH the bridge_child cascade path AND a direct-search path lands in `recommendable_concepts[]` **twice**, because the two paths produce different `concept_key` strings:

- Bridge cascade target: `["…","BD50.3Z&XA01A6","HEAD"]`
- Direct match:          `["…","BD50.3Z&XA01A6",null]`

In `buildV2RecommendationPayload` ([src/components/map-projects/MapProject.jsx](src/components/map-projects/MapProject.jsx#L3703)), `defsByKey` keys on the full opaque key, so these are treated as distinct entries and the evidence from the two paths never converges. The LLM then sees the same concept twice with split evidence and inflated payload size.

## Root cause

`cascadeTargetToConceptDefinition` in [src/components/map-projects/normalizers.js](src/components/map-projects/normalizers.js#L156-L190) (named `toBridgeChildConceptDefinition` in the master plan) delegates to `resolveReference`, which inherits `projectContext.target_repo.version` ([normalizers.js:97-98](src/components/map-projects/normalizers.js#L97-L98)). In production projects pinned at "HEAD", that string becomes the version slot on the cascade-target reference.

Direct-algorithm matches against the same target repo arrive without a version slot (their `reference_source` chain either resolves a real version or `null`, depending on the algo). The two paths therefore disagree on what should be the same concept identity.

## Fix

In `cascadeTargetToConceptDefinition`, drop the version slot on the resolved reference unconditionally. `"HEAD"` isn't a stable version identifier — it's the project pin, not a per-concept version — and the cascade-target stub is always going to be re-resolved via `ensureLoaded` anyway. Aligning both paths on no-version is cleaner than pinning both to a sentinel.

## Verification

- Unit test in [src/components/map-projects/__tests__/normalizers.test.js](src/components/map-projects/__tests__/normalizers.test.js): with `projectContext.target_repo.version = 'HEAD'`, a bridge cascade-target ConceptDefinition's `key` does not include "HEAD"; it matches the key produced by `makeConceptKey({url, code})` for a direct match of the same (url, code).
- Same-(url, code) bridge_child + direct-path produces ONE entry in the v2 payload's `recommendable_concepts` with `evidence.length === 2`. (PR3-G handles the orthogonal merge-order bug; both need to land for full convergence.)

## Context

Part of the post-PR2b OCL Mapper unification close-out (PR3-X — auditability and visibility first). PR3-X.1 (audit-trail snapshot) shipped in [ocl-ai-assistant#139](https://github.com/OpenConceptLab/ocl-ai-assistant/pull/139). This ticket is the oclmap half of PR3-X. See `ocl-online-docs/mapper/unified-mapper-model.md` -> "PR3 Implementation Plan" -> "PR3-X.2" for the slice spec.

Refs OpenConceptLab/ocl_issues#2337 (umbrella).

paynejd:
Closes OpenConceptLab/ocl_issues#2520. Refs OpenConceptLab/ocl_issues#2337.

## Summary

`cascadeTargetToConceptDefinition` (called `toBridgeChildConceptDefinition` in the master plan) inherits `projectContext.target_repo.version` onto the cascade-target's reference via `resolveReference`. Production projects pinned at `"HEAD"` end up with `["…","HEAD"]` in the version slot of the cascade-target's `concept_key`. Direct fixed-canonical matches against the same (url, code) arrive with `null` in that slot, so the two paths produce different keys, `defsByKey` in `buildV2RecommendationPayload` treats them as distinct entries, the LLM sees the same concept twice with split evidence, and the v2 prompt payload bloats.

This PR drops the inherited version slot on cascade-target references unconditionally. `"HEAD"` is the project repo pin, not a per-concept version, and the cascade-target stub is re-resolved via `ensureLoaded` anyway. Aligning both paths on no-version is cleaner than pinning both to a sentinel.

## Diff

- [src/components/map-projects/normalizers.js](src/components/map-projects/normalizers.js#L156-L190) — in `cascadeTargetToConceptDefinition`, build the final `reference` as `{url, code}` from the resolved reference, discarding any inherited `version`.
- [src/components/map-projects/__tests__/normalizers.test.js](src/components/map-projects/__tests__/normalizers.test.js) — three new tests under a "PR3-X.2: bridge cascade target version-slot dedup" section.

## Tests

89/89 pass. The new tests cover:

1. **Cascade target reference does not inherit target_repo.version, even when set.** Asserts the surgical fix.
2. **Bridge cascade target key matches direct fixed-canonical match key under pinned target_repo.version="HEAD".** Authentic prod scenario — `ocl-bridge` cascade target vs `ocl-scispacy-loinc` (fixed canonical) on the same `(http://loinc.org, 49494-3)`. Pre-fix: different keys; post-fix: same key.
3. **Bridge cascade target collides with direct fixed-canonical match in `defsByKey` dedup.** Proxies the dedup logic in `buildV2RecommendationPayload`: one bridge_child + one standard candidate converge on the same `concept_key`, exactly the "evidence.length === 2" outcome the master plan calls for. `buildV2RecommendationPayload` itself isn't unit-testable as written (closure inside the React component); this is the closest layer below that's pure.

eslint clean.

## Why two slices for X.1 + X.2

X.1 (audit trail in ocl-ai-assistant, [#139](https://github.com/OpenConceptLab/ocl-ai-assistant/pull/139)) and X.2 (this one) are both visibility/auditability fixes that need to land before PR3-D1's payload trim so we can read the before/after diff cleanly in Django admin.

## Coordination with PR3-G

PR3-G fixes the orthogonal "merge didn't happen because one entry was thinner" bug (`mergeIntoRowMatchState` richer-wins). This PR fixes "merge didn't happen because the keys never collided" (key construction). Both need to land for the LLM to see a single deduplicated entry per concept with all evidence converged.

## Context

Part of the post-PR2b OCL Mapper unification close-out. See `ocl-online-docs/mapper/unified-mapper-model.md` -> "PR3 Implementation Plan" -> "PR3-X.2" for the slice spec.

## Test plan

- [ ] Merge.
- [ ] In staging, open a project with `target_repo.version` pinned (e.g., LOINC at "HEAD") and a row that produces both a bridge cascade hit and a direct fixed-canonical match on the same target concept. Trigger AI Assistant.
- [ ] Inspect the staging POST body to `/prompts/.../invoke/`: `recommendable_concepts[]` has ONE entry for that concept with `evidence.length === 2` (one `bridge_child` + one direct).
- [ ] Combined with X.1 already shipped, the audit-trail's `input_variables` vs `processed_variables` shows the dedup.

### DIFF
--- a/src/components/map-projects/__tests__/normalizers.test.js
+++ b/src/components/map-projects/__tests__/normalizers.test.js
@@ -610,6 +610,118 @@ test('bridge result with multiple cascade targets fans out 1 + N candidates', ()
   }
 })
 
+// ---------- PR3-X.2: bridge cascade target version-slot dedup ----------
+//
+// Regression guard for OpenConceptLab/ocl_issues#2520. When the project's
+// target_repo carries a `version` field (commonly "HEAD" in production
+// projects pinned at the head of the source), the cascade target previously
+// inherited that version onto its reference. Direct-search matches against
+// the same (url, code) arrive without a version slot, so the two paths
+// produced different concept_keys (`[…,"HEAD"]` vs `[…,null]`) and
+// `defsByKey` / `recommendable_concepts` saw two entries instead of one.
+//
+// Fix: cascadeTargetToConceptDefinition drops the inherited version slot
+// unconditionally. These tests assert that fix at the normalizer layer; the
+// downstream effect (one entry in `recommendable_concepts` with two pieces of
+// evidence) follows from `buildV2RecommendationPayload`'s pre-existing
+// key-based dedup.
+
+test('cascade target reference does not inherit target_repo.version, even when set', () => {
+  const out = normalizeAlgoResult(bridgeResult_CIEL_to_LOINC, {
+    algorithmId: 'ocl-bridge',
+    algorithmConfig: oclBridgeAlgo,
+    algorithmResponseId: 'ar-pinned',
+    projectContext: {
+      ...projectContext,
+      target_repo: { ...projectContext.target_repo, version: 'HEAD' }
+    }
+  })
+
+  const targetDef = out.concept_definitions.find(d => d.reference.code === '49494-3')
+  assert.ok(targetDef, 'cascade target ConceptDefinition was emitted')
+  assert.equal(targetDef.reference.version, undefined,
+    'cascade-target reference must NOT carry the inherited target_repo.version sentinel')
+  assert.deepEqual(targetDef.reference, { url: 'http://loinc.org', code: '49494-3' })
+})
+
+test('bridge cascade target key matches direct fixed-canonical match key under pinned target_repo.version', () => {
+  // Same (url, code) — '49494-3' on http://loinc.org — arriving via two paths:
+  // (a) ocl-bridge cascade target on a CIEL bridge result (resolves via
+  //     target_repo, so pre-fix inherited target_repo.version="HEAD")
+  // (b) ocl-scispacy direct match (resolves via fixed canonical, no version)
+  // Pre-fix the keys diverged (`[…,"HEAD"]` vs `[…,null]`) and the LLM saw
+  // the concept twice with split evidence. Post-fix the cascade target drops
+  // the inherited version, so both paths key on the same canonical (url, code).
+  const pinnedCtx = {
+    ...projectContext,
+    target_repo: { ...projectContext.target_repo, version: 'HEAD' }
+  }
+
+  const bridgeOut = normalizeAlgoResult(bridgeResult_CIEL_to_LOINC, {
+    algorithmId: 'ocl-bridge',
+    algorithmConfig: oclBridgeAlgo,
+    algorithmResponseId: 'ar-bridge',
+    projectContext: pinnedCtx
+  })
+  // Scispacy fixture has the same code (49494-3) on the same canonical
+  // (http://loinc.org via reference_source='fixed').
+  const directOut = normalizeAlgoResult(scispacyResult_partial, {
+    algorithmId: 'ocl-scispacy-loinc',
+    algorithmConfig: oclScispacyAlgo,
+    algorithmResponseId: 'ar-direct',
+    projectContext: pinnedCtx
+  })
+
+  const cascadeTargetDef = bridgeOut.concept_definitions.find(d => d.reference.code === '49494-3')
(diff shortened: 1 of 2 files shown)


paynejd:
Superseded by #23.

After review, the fix direction in this PR was incorrect for the common case. Summary of the corrected analysis:

A mapping project pins to ONE explicit `target_repo` version (typically a released version like LOINC `2.81`; HEAD is allowed but uncommon and discouraged). Concept identity for dedup must be anchored to that pin across all algorithm paths.

Pre-PR-#22 partition under a `2.81`-pinned project:
- `{ocl-search, ocl-semantic, ocl-bridge cascade target}` all keyed `[…,"2.81"]` — correct
- `{ocl-scispacy}` keyed `[…,null]` — the actual outlier

PR #22 fixed the split by *dropping* the version slot on bridge_child cascade target, making it align with scispacy at `null`. That worked for the HEAD-pinned reproduction in #2520 but for the common released-version case it pulled `bridge_child` away from its correct alignment with search/semantic. Also subtle: with no version on the reference, `ensureLoaded` would POST `{url}` only to `$resolveReference`, which resolves to "latest released or HEAD if no other versions exist" — NOT to the project pin. The bridge_child stub would load concept details from the wrong version of the repo.

The corrected fix in #23 aligns scispacy (the actual outlier) UP to the project pin by extending `resolveReference`'s `fixed` branch to inherit `target_repo.version` when the algorithm's `canonical_url` matches `projectContext.target_repo.canonical_url`. All four algorithm paths then produce identical concept_keys, anchored to whatever version the project pins. `ensureLoaded` automatically carries the pin through `$resolveReference` via the existing `reference.version ? {url, version} : {url}` logic.

Two follow-up tickets filed:
- OpenConceptLab/ocl_issues#2521 — surface algorithm-claimed version divergence to user (for future case where a `$cascade` response or external algo claims a different version than the project pin).
- OpenConceptLab/ocl_issues#2522 — require explicit `target_repo.version` on the project config form; remove silent `'HEAD'` and `undefined` defaults.

See #23 for the replacement.

PR #29, opened 2026-05-24
OpenConceptLab/ocl_issues#2536 | PR3-G instrumentation: trace conceptCache writes

paynejd:
## Summary

Diagnostic work for #2536 (PR3-G — `display_name: null` on bridge_child cascade targets in the AI payload despite the lookup repo returning 200).

This PR is **not the fix** — it carries the investigation infrastructure that determines where the bug lives.

## What lands here

**1. Pure-logic reproducer** (`src/components/map-projects/__tests__/pr3g-cache-bug-repro.test.js` + 3 fixtures)

Replays the run-2 bridge→lookup→second-bridge→AI-payload sequence using the real `normalizeAlgorithmInvocation` + `buildRecommendableConceptEntry`, pure-function mirrors of `mergeIntoRowMatchState`'s cache branch and `writeConceptCachePatch`, and real captured fixture data from the 2026-05-21 ICD-11 IMO run.

**Result: both assertions PASS (142/142 tests).** That's a definitive negative result — the rank-guarded merge, lookup write-back, and AI-payload builder are all logically correct given a well-ordered sequence. **The PR3-G bug must therefore live in React lifecycle, not pure logic.**

**2. Console-log instrumentation** (`src/components/map-projects/__debug/conceptCacheTrace.js` + 5 trace call sites in `MapProject.jsx`)

localStorage-gated; zero impact when disabled. Captures every conceptCache write/read for a target concept_key, with `source`, `prev` def, `next` def. Five trace points cover: `mergeIntoRowMatchState` accept + reject branches, `writeConceptCachePatch`, the useEffect at line 1738 (suspect A), the legacy URL-keyed setConceptCache at line 2407 (suspect B), and `buildV2RecommendationPayload` read.

Useful for confirming the React-lifecycle root cause once we set up either an integration test or browser automation.

Enable in browser devtools:

```js
localStorage.setItem('MAPPER_DEBUG_CONCEPT_CODE', 'QC01.2')
// or by exact concept_key:
localStorage.setItem('MAPPER_DEBUG_CONCEPT_KEY',
  '["http://id.who.int/icd/release/11/mms","QC01.2","HEAD"]')
```

## Why this matters

Before this PR's reproducer, we couldn't tell whether the bug was in pure logic or React lifecycle without manual browser testing. After this PR:

- **Pure logic exonerated.** Don't waste future cycles auditing the merge/write paths.
- **Lifecycle suspects narrowed** to the two paths in #2536: the `useEffect` at MapProject.jsx:1738 and the legacy URL-keyed setConceptCache at MapProject.jsx:2407.
- **Fixtures + sequence harness are reusable** — once we know the actual fix, the test extends naturally to a regression guard (likely by mounting MapProject in a real React lifecycle via RTL + happy-dom).

## Test plan

- [x] `npm test` — 142/142 pass (140 existing + 2 new). Both PR3-G assertions pass, confirming pure logic is sound.
- [x] `npm run eslint` — clean.
- [ ] Browser-side confirmation of root cause (suspect A vs B vs other) using the instrumentation — manual or via follow-up integration test.

## What this PR does NOT do

- Does not ship the fix for #2536. The fix lands as a follow-up PR once the lifecycle root cause is confirmed.
- Does not add React Testing Library or other test-framework deps. The pure-logic reproducer uses only the existing `node:test` runner. RTL adoption is a separate test-infra decision (raised on #2536 / #29 thread).

## Related

- #2536 — the bug
- [`ocl-online-docs/mapper/unified-mapper-model.md` PR3-G](https://github.com/OpenConceptLab/ocl-online-docs/blob/main/mapper/unified-mapper-model.md#pr3-g--fix-display_name-null-on-bridge_child-cascade-targets-lookup-ok-case) — design

### DIFF
--- a/src/components/map-projects/MapProject.jsx
+++ b/src/components/map-projects/MapProject.jsx
@@ -100,6 +100,8 @@ import { useAlgos, CONCEPT_IDENTITY_BY_TYPE, ensureConceptIdentity } from './alg
 import AutoMatchDialog from './AutoMatchDialog'
 import { DEFAULT_ENCODER_MODEL } from './rerankerModels'
 import { normalizeAlgorithmInvocation, lookupStatusRank, normalizeLegacyAllCandidates, buildRecommendableConceptEntry, stripConstantClassAndDatatype } from './normalizers'
+// Debug-only — remove with OpenConceptLab/ocl_issues#2536 fix
+import { traceConceptCacheWrite, traceConceptCacheRead } from './__debug/conceptCacheTrace'
 import { parseConceptKey } from './conceptKey'
 import { buildQualityRowViews, conceptForMapping, resolveAICandidateID } from './viewBuilders.js'
 
@@ -377,8 +379,14 @@ const MapProject = () => {
     concept_definitions.forEach(def => {
       const existing = nextCache[def.key]
       if(!existing || lookupStatusRank(def.lookup_status) > lookupStatusRank(existing.lookup_status)) {
+        // Debug — #2536
+        traceConceptCacheWrite(def.key, 'mergeIntoRowMatchState/accept', existing, def, { rowIndex, algorithmId: incomingAlgoId })
         nextCache[def.key] = def
         cacheChanged = true
+      } else {
+        // Debug — #2536: log rejections too, so we can see when the rank guard
+        // blocks a thinner write that might have been desired.
+        traceConceptCacheWrite(def.key, 'mergeIntoRowMatchState/reject(rank)', existing, def, { rowIndex, algorithmId: incomingAlgoId })
       }
     })
     if(cacheChanged) {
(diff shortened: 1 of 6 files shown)


paynejd:
Closing without merging — superseded by [#30](https://github.com/OpenConceptLab/oclmap/pull/30) (RTL + happy-dom harness + failing reproducer) and [#31](https://github.com/OpenConceptLab/oclmap/pull/31) (the actual fix for #2536).

This PR's two artifacts both did their diagnostic job upstream of those:

- The **pure-logic reproducer** (`__tests__/pr3g-cache-bug-repro.test.js`) confirmed the bug isn't in the merge/write logic, which directly motivated standing up the React-lifecycle harness in #30. The integration test in #30 now provides broader coverage (catches the actual lifecycle bug), so the pure-logic test doesn't add forward value.
- The **console-trace instrumentation** (`__debug/conceptCacheTrace.js` + 5 trace call sites in `MapProject.jsx`) would have been useful for browser-side root-causing if #30 hadn't reproduced the bug. It did, so the instrumentation isn't needed.

Neither belongs in production code now that the root cause is known and the fix is targeted. The historical investigation breadcrumb stays intact via #2536's discussion + the trail from #30 → #31.

PR #30, opened 2026-05-24
chore(test): add React component test harness (RTL + happy-dom); ships failing PR3-G repro

paynejd:
## Summary

Adds `@testing-library/react` + `happy-dom` as a **coexisting** layer on top of the existing `node:test` suite. Pure-function unit tests under `__tests__/` are unchanged. New React component tests live under `__integration__/` and run via `npm run test:integration`.

Also ships a failing test that reproduces #2536 (PR3-G) — see "Diagnostic finding" below. The fix lands in a follow-up PR.

## Why now

PR3-G (#2536) requires a React-lifecycle reproducer that pure functions can't provide. A pure-logic reproducer ([PR #29](https://github.com/OpenConceptLab/oclmap/pull/29)) already ruled out the merge/write logic — the bug must be in React lifecycle.

This harness also unblocks PR3-C (schema-v2 save/load round-trip — the unified plan lists this as needing real React lifecycle for the legacy load path) and any future React-lifecycle bug.

## Design decisions

| Decision | Why |
|---|---|
| **Coexist with `node:test`** (not migrate to Vitest) | Keeps PR1's "zero new framework deps" spirit. Vitest stays a future option — the conventions in `__integration__/README.md` keep migration to a ~30-min codemod (imports + assertions). |
| **happy-dom over jsdom** | ~5× faster on startup, covers everything React 18 + our patterns need. If a future test needs a jsdom-specific API, swap per-test or globally — RTL API is identical. |
| **`React.createElement` in test files instead of JSX** | Avoids a transform step. Tests run via plain `node --test`. If JSX becomes painful, add `tsx` or migrate to Vitest. |
| **`__integration__/` directory** (not `__tests__/integration/`) | Existing `test` glob (`'src/**/__tests__/**/*.test.js'`) is untouched. The two suites are visibly separated in the source tree. |

## What lands

**Build / config:**
- `package.json`: + `@testing-library/react@^16.1`, `@testing-library/dom@^10.4`, `@happy-dom/global-registrator@^15.11` (devDeps). + `test:integration` and `test:all` scripts.
- `test-integration-setup.js` (repo root): 20 lines. Registers happy-dom globals + `IS_REACT_ACT_ENVIRONMENT`. Loaded once per integration test process via `node --import`.

**Conventions doc:**
- `src/components/map-projects/__integration__/README.md`: when to write a unit vs integration test, Vitest-compatibility rules, why happy-dom + `React.createElement`.

**Tests:**
- `_harness-smoke.test.js` — 3 tests verifying the harness works (globals registered, RTL render + query works, cleanup works between tests).
- `pr3g-display-name-loss.test.js` — 3 tests. Mounts a minimal harness mirroring MapProject.jsx's cache pattern: `useState` + `useRef` + the `useEffect` at L1738 + `writeConceptCachePatch` (L3308) + the rank-guarded merge from `mergeIntoRowMatchState` (L377-395) + the legacy URL-keyed `setConceptCache` at L2472. Drives the harness through the suspect interaction from #2536.

## Diagnostic finding

Running locally:

```text
npm test                   → 140/140 pass  (existing pure-function suite, unchanged)
npm run test:integration   → 5/6 pass, 1 FAILS as #2536 predicts

  ✔  PR3-G/1 baseline: merge stub then lookup-patch leaves enriched entry in ref
  ✔  PR3-G/2 re-merge: a second bridge merge cannot overwrite the enriched entry
  ✖  PR3-G/3 STALE CLOSURE: legacy URL-keyed setConceptCache wipes lookup enrichment

  AssertionError: PR3-G: lookup enrichment must survive a deferred stale-closure setConceptCache (#2536)
    + null
    - 'Need for immunization against rabies'
```

The failing test reproduces the bug deterministically: a `setConceptCache` call that uses a closure-captured stale `conceptCache` (the pattern at MapProject.jsx:2472, `setConceptCache({...conceptCache, [url]: res})`) overwrites React state with a stale snapshot. The `useEffect` at MapProject.jsx:1738 then copies that stale state back into `conceptCacheRef`, wiping out any synchronous enrichment that `writeConceptCachePatch` had made in between.

The two suspects in #2536's investigation are **both load-bearing** — neither alone causes the bug. The (A)+(B) interaction is the bug.

## Why we ship the failing test now

This PR deliberately ships the failing test. It's the diagnostic data the next PR (the fix) needs. The fix PR will:
- Make `pr3g-display-name-loss/3` pass (either by removing the L1738 useEffect, replacing the stale-closure setConceptCache at L2472 with a ref-based write, or both)
- Add follow-up tests asserting all 5 enriched fields (display_name, names, concept_class, datatype, rerank_score) post-fix
- Close #2536

## Test plan

- [x] `npm test` — 140/140 pass (existing suite unchanged)
- [x] `npm run test:integration` — 5/6 pass, 1 fails as documented (the diagnostic)
- [x] `npm run eslint` — clean
- [ ] CI does not currently run `test:integration`. Wiring it into CI is a follow-up — for now, the harness exists and is locally runnable.

## Related

- #2536 — bug ticket (full investigation + scope clarification)
- #29 — instrumentation PR (the pure-logic reproducer + browser-side console traces). Will be closed or merged depending on whether the fix PR's regression test fully supersedes its value.
- [`ocl-online-docs/mapper/unified-mapper-model.md` PR3-G](https://github.com/OpenConceptLab/ocl-online-docs/blob/main/mapper/unified-mapper-model.md#pr3-g--fix-display_name-null-on-bridge_child-cascade-targets-lookup-ok-case) — design doc

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,6 +42,9 @@
         "@babel/preset-env": "^7.26.9",
         "@babel/preset-react": "^7.26.3",
         "@babel/runtime": "^7.26.10",
+        "@happy-dom/global-registrator": "^15.11.7",
+        "@testing-library/dom": "^10.4.0",
+        "@testing-library/react": "^16.1.0",
         "babel-eslint": "^10.1.0",
         "babel-loader": "^8.4.1",
         "copy-webpack-plugin": "^4.6.0",
@@ -2085,6 +2088,19 @@
       "integrity": "sha512-RiB/yIh78pcIxl6lLMG0CgBXAZ2Y0eVHqMPYugu+9U0AeT6YBeiJpf7lbdJNIugFP5SIjwNRgo4DhR1Qxi26Gg==",
       "license": "MIT"
     },
+    "node_modules/@happy-dom/global-registrator": {
+      "version": "15.11.7",
+      "resolved": "https://registry.npmjs.org/@happy-dom/global-registrator/-/global-registrator-15.11.7.tgz",
+      "integrity": "sha512-mfOoUlIw8VBiJYPrl5RZfMzkXC/z7gbSpi2ecycrj/gRWLq2CMV+Q+0G+JPjeOmuNFgg0skEIzkVFzVYFP6URw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "happy-dom": "^15.11.7"
+      },
+      "engines": {
+        "node": ">=18.0.0"
+      }
+    },
     "node_modules/@humanwhocodes/config-array": {
       "version": "0.13.0",
       "resolved": "https://registry.npmjs.org/@humanwhocodes/config-array/-/config-array-0.13.0.tgz",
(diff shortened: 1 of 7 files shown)


snyaggarwal (commented):
This looks good.
need to include in .github/workflows/build.yaml to run `npm test:all`

snyaggarwal (approved):

PR #42, opened 2026-06-04
Mapper UI auto-match attribution: AutomatchRun lifecycle + propagation headers (ocl_online#105 Phase 5)

paynejd:
Phase 5 (the final implementation phase) of **OpenConceptLab/ocl_online#105** — retrospective process attribution. Closes the chain: the Mapper UI is the writer that creates the run record and stamps the attribution headers the already-deployed middleware reads.

Sub-ticket: **OpenConceptLab/ocl_online#115**.

## What this does
- **Creates an `AutomatchRun`** at auto-match start — `POST <project>/auto-match-runs/` with `intended_rows`, `trigger_source='ui-auto-match'`, and a `config_snapshot` (algorithms, encoder, score thresholds, filters, AI template/model). Captures the returned `id`.
- **Stamps attribution headers on every per-row backend call** — `X-OCL-Request-Source` (discrete) + `X-OCL-Event-Metadata` (single-line, string-valued JSON bag), across `$match`, `$rerank`, **and the read-side** `$resolveReference` + concept-detail GET, plus the AI `$invoke` (keeping the existing discrete idempotency header and adding the discrete prompt-template-key/version headers). `$match` is the one batched call → `row_indices` + `batch_size`; the rest carry a scalar `row_index`.
- **PATCHes the run to a terminal status** (`completed` / `partial` / `failed` / `cancelled`) with `completed_rows` / `failed_rows` at run end — wrapped in `try/finally` so cancel or an unexpected throw still records the run.
- **Manual (non-run) reuse of the same call sites** (panel open, single rerun) emits `request_source=mapper-ui-manual` for free, so cost-by-surface works for manual usage too.

## Key files
- `src/services/attribution.js` (new) — pure `buildAttributionHeaders` (string-valued scalars, `row_indices` int array, null-key omission, manual fallback) + `buildConfigSnapshot` (strips non-serializable algo fields). 13 unit tests in `src/services/__tests__/attribution.test.js`.
- `src/components/map-projects/MapProject.jsx` — `automatchRunRef` + `createAutomatchRun` / `completeAutomatchRun` + the 6 call-site one-liners. Run creation **degrades gracefully** (a failed create never blocks matching).

## Contract verification (against the deployed backends)
- All upstream phases merged + deployed; `/map-projects/<id>/auto-match-runs/` and `/auto-match-runs/<id>/` resolve in prod (403, not 404).
- Live prod CORS preflight on oclapi2 `$match` returns `access-control-allow-headers` including `X-OCL-REQUEST-SOURCE`, `X-OCL-EVENT-METADATA`, `X-OCL-PROMPT-TEMPLATE-KEY/VERSION`, `X-OCL-REQUEST-IDEMPOTENCY-KEY`; ai-assistant settings list the same. Scalars are sent as JSON **strings** because the middleware filters compare `event_metadata->>'key'` (text).
- `eslint` clean; **158 tests pass**; the running webpack dev bundle compiles with the new code.

## Scope notes
- **Scispacy is now attributed** (correcting an earlier wrong assumption — it is NOT a dark lambda). `ocl-scispacy-loinc` runs its own `AnalyticsMiddleware` and records ~2.7k `$match` rows/30d in `api_transactions`; its `$match` is instrumented here, with the CORS + emitter forward-list change in the companion **OpenConceptLab/ocl-scispacy-loinc#3**.
- **Custom `$match` services remain unrecorded by design** — a custom algo POSTs to `algo.url` (external) with no OCL middleware, so there is no `api_transactions` row. Server-side recording is inherently first-party-only; capturing custom/external `$match` at 100% needs an oclmap-side per-call recorder, tracked as a future enhancement under #105. The AutomatchRun aggregate (config_snapshot + completed/failed counts) already covers custom algos at the run level.
- **Bridge `$match`** attribution is supplied here but applied by the **react-bridge-match** premium component — see OpenConceptLab/react-bridge-match#2. Passing the new arg is harmless against the current bundle; full bridge coverage needs that bundle rebuilt + redeployed at `BRIDGE_MATCH_URL`.

## Remaining for #105 to close
1. Merge the react-bridge-match PR (**merged**) + redeploy the hosted bridge bundle; merge **ocl-scispacy-loinc#3** + deploy scispacy.
2. Optional end-to-end smoke (a real auto-match run writes a live `AutomatchRun` row + attributed `api_transactions`/`prompt_executions`).

### DIFF
--- a/src/components/map-projects/MapProject.jsx
+++ b/src/components/map-projects/MapProject.jsx
@@ -68,6 +68,7 @@ import pick from 'lodash/pick'
 import { OperationsContext } from '../app/LayoutContext';
 
 import APIService, { isTransientNetworkError, retryWithBackoff } from '../../services/APIService';
+import { buildAttributionHeaders, buildConfigSnapshot } from '../../services/attribution'
 import { highlightTexts, dropVersion, getCurrentUser, hasAuthGroup, downloadObject, currentUserToken } from '../../common/utils';
 import { WHITE, SURFACE_COLORS } from '../../common/colors';
 
@@ -199,6 +200,10 @@ const MapProject = () => {
   const abortRef = React.useRef(false);
   const isBulkMatchRunningRef = React.useRef(false);
   const bulkMatchAlgoIdsRef = React.useRef([]);
+  // ocl_online#105 Phase 5: the active AutomatchRun ({id, algoIds}) or null.
+  // A ref so every per-row backend call reads the run id synchronously without
+  // prop-drilling or re-renders; null outside a run → 'mapper-ui-manual'.
+  const automatchRunRef = React.useRef(null);
 
   const [row, setRow] = React.useState(false)
   const [loadingMatches, setLoadingMatches] = React.useState(false)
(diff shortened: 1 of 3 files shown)


src/services/attribution.js
```diff
@@ -0,0 +1,121 @@
...
+  if(isPresent(algorithmId)) meta.algorithm_id = String(algorithmId)
+  if(isPresent(clientAttemptN)) meta.client_attempt_n = String(clientAttemptN)
+
+  return {
+    [REQUEST_SOURCE_HEADER]: requestSource,
+    [EVENT_METADATA_HEADER]: JSON.stringify(meta),
+  }
+}
+
+const isFunction = value => typeof value === 'function'
```
snyaggarwal: import from [lodash#isFunction](https://lodash.com/docs/4.18.1#isFunction)

src/components/map-projects/MapProject.jsx
```diff
@@ -1294,6 +1299,77 @@ const MapProject = () => {
...
+    const stages = rowStageRef.current || []
+    const algoIds = run.algoIds?.length ? run.algoIds : map(selectedAlgos, 'id')
+    let failed = 0
+    forEach(intendedRows, _row => {
+      const rowStage = stages[_row.__index] || {}
+      const attempted = algoIds.map(id => rowStage[id]).filter(s => s !== undefined && s !== -1)
+      if(attempted.length && attempted.every(s => s === -2)) failed += 1
+    })
+    const total = intendedRows.length
+    const completed = total - failed
```
snyaggarwal: Cancelled runs will over-report completed_rows. this treats every non-failed row as completed but rows that were never attempted after abortRef.current flips stay at -1/undefined and are still counted as completed.

src/components/map-projects/MapProject.jsx
```diff
@@ -1294,6 +1299,77 @@ const MapProject = () => {
       APIService.new().overrideURL(project.url).appendToUrl('logs/').post({logs: {row_logs: logs, project_logs: newLogs}}).then(() => {})
   }
 
+  // ── ocl_online#105 Phase 5: AutomatchRun attribution ──────────────────────
+  // Build the X-OCL-Request-Source + X-OCL-Event-Metadata headers for a single
+  // per-row backend call. Reads the active run id from the ref, so the same
+  // call sites used outside a run emit request_source='mapper-ui-manual'.
+  const attrHeaders = ({rowIndex, rowIndices, batchSize, algorithmId, clientAttemptN} = {}) =>
+    buildAttributionHeaders({
```
snyaggarwal, line 1313: attrHeaders() reads a single global automatchRunRef, so any shared/manual call site invoked while an auto-match is in progress will be stamped as automatch. That includes the read-side $resolveReference / concept GET paths used when a user opens or refreshes a row during a run. The fallback to mapper-ui-manual only works when no run is active, not when manual and run traffic overlap.

snyaggarwal (changes requested):

src/services/attribution.js
paynejd: Done in d2db303 — now `import isFunction from 'lodash/isFunction.js'` (the `.js` is required so the node:test ESM runner resolves it; webpack handles it either way).

src/components/map-projects/MapProject.jsx
paynejd: Fixed in d2db303. Extracted the counting into a pure `summarizeRunCompletion()` (attribution.js): a row is **completed** only when an attempted algo finished (stage 1), **failed** when every attempted algo failed (-2); rows that never ran (-1/undefined — exactly the post-abort case you flagged) or are still in flight (0) count as **neither**, so a cancelled run no longer over-reports `completed_rows`. Added 7 unit tests incl. an explicit cancel-with-unreached-rows case.

src/components/map-projects/MapProject.jsx
paynejd: Good catch — fixed in d2db303 by making attribution key on explicit call **origin** instead of the ambient ref. `attrHeaders` now takes `isRunTraffic` and only stamps the run id when true. Pure-bulk `$match` passes true; `$rerank`/bridge/`$invoke` pass their existing `isBulk`; and for the shared read-side I thread it through the bulk merges (`mergeIntoRowMatchState({isRunTraffic:true})`) into `ensureLoaded` → `$resolveReference`/concept GET. Every manual call site defaults false, so a panel open / single rerun / per-row AI that overlaps a running auto-match is now `mapper-ui-manual`.

snyaggarwal (approved):

PR #43, opened 2026-06-04
OpenConceptLab/ocl_issues#2559 | Ability to lock bridge repo version

### LINKED ISSUE #2559: Bridge Algo fails when latest Bridge Repo version is released and not (yet!) vectorised and project was already using prev latest
**Problem:**
Bridge Algo definition doesn't talk about bridge repo version, and that gives freedom to algo implementation to always use `latest` version. That `latest` version may be a different/new version that just got released. This causes issues like:
1. MapProject evaluated with prev latest bridge repo version wont work on re-evalaution and a silent console log is added that this bridge repo version is not yet vectorised.
2. MapProject evaluated with prev latest (vectorised) bridge repo version may see new updates which are not explicitly intentional.

**Solution:**
- [x] Add `target_repo_version` in bridge algo with `target_repo_url` with default value as `latest`
- [x] Add `version` selector in Mapper UI defaulted to resolved (latest) and saved against the project
- [ ] Migrate existing projects to have saved "target_repo_version" to resolved `latest` (by date?)
- [x] Drop `latest` from bridge algo call and use `target_repo_version` always
- [x] Also it should log which bridge repo version was used

snyaggarwal:
## Linked Issue

Refs: OpenconceptLab/ocl_issues#2559

Companion - https://github.com/OpenConceptLab/react-bridge-match/pull/3


src/components/map-projects/MapProject.jsx
```diff
@@ -1471,6 +1473,10 @@ const MapProject = () => {
     })
   }
 
+  const getAlgoLogExtras = algo => {
+    return {algo: algo.id, repo_url: algo.target_repo_url, ...(algo.target_repo_version ? {repo_version: algo.target_repo_version} : {})}
+  }
```
paynejd, line 1513: `getAlgoLogExtras` dereferences `algo.id` unconditionally, but `getAlgoDef` can return `undefined` (see its `if(!algo) return algo` at line ~2546). In the decision-view path this helper is now called eagerly at the top of `onResponse` (`const logExtras = getAlgoLogExtras(getAlgoDef(algoId))`), *before* the `response?.detail` failure branch — so an unresolved algo would throw a `TypeError` and break the very `algo_failed` logging/alert path. The old code (`{algo: algoId}`) couldn't fail this way. A cheap guard:

```suggestion
  const getAlgoLogExtras = algo => {
    if(!algo) return {}
    return {algo: algo.id, repo_url: algo.target_repo_url, ...(algo.target_repo_version ? {repo_version: algo.target_repo_version} : {})}
  }
```

(Also `let _algo` at line ~2655 can be `const`.)

src/components/map-projects/MultiAlgoSelector.jsx
```diff
@@ -49,6 +55,24 @@ const fetchBridgeCanonical = (url) => {
...
+  if(!baseUrl) return Promise.resolve([])
+  if(bridgeVersionsCache.has(baseUrl)) return bridgeVersionsCache.get(baseUrl)
+  const promise = APIService.new().overrideURL(baseUrl).appendToUrl('versions/').get()
+    .then(r => r?.data || [])
+    .catch(() => [])
+  bridgeVersionsCache.set(baseUrl, promise)
+  return promise
+}
+
+const getLatestReleasedBridgeVersion = (versions) => {
```
paynejd: Two divergent notions of "latest" coexist: this local `getLatestReleasedBridgeVersion` returns the first `released` version **ignoring LLM support**, while `getDefaultBridgeRepoVersion` (in projectTargetRepo.js, used to auto-select the default) returns the first `released` **LLM** version. The "Latest" chip uses the former, the actual selection uses the latter — so when the newest release isn't vectorised, the UI labels a *disabled* option "Latest" while silently defaulting the selection to an older version. Consider consolidating to one definition (or labeling the chip on the selectable latest) to avoid the confusing mismatch. This local helper is also untested vs. the extracted one.

src/components/map-projects/projectTargetRepo.js
```diff
@@ -17,6 +19,17 @@ export const getDefaultTargetRepoVersion = versions => {
   return false
 }
 
+export const getDefaultBridgeRepoVersion = versions => {
```
paynejd, line 22: This relies on `versions` being newest-first (`.find(released && llm)` == "latest released llm"). The sibling `getDefaultTargetRepoVersion` deliberately avoids any ordering assumption (it only auto-picks when exactly one version is released). Baking in the API's ordering contract is fine if it's guaranteed — worth a one-line comment noting the assumption so it isn't silently broken if the endpoint's ordering ever changes.

src/components/repos/RepoVersionSearchAutocomplete.jsx
```diff
@@ -1,9 +1,24 @@
...
+  label,
+  id,
+  required,
+  size,
+  sx,
+  value,
+  error,
+  helperText,
+  getOptionDisabled,
+  renderOptionLabel,
```
paynejd: `renderOptionLabel` is a new prop but no caller passes it (only `getOptionDisabled` / `getOptionRightText` are used by MultiAlgoSelector; ConfigurationForm passes neither). It's harmless (guarded ternary at line ~53 falls back to `getLabel`), but it's dead API surface — suggest dropping both the destructure here and the reference at line ~53 unless there's a planned use.

src/components/map-projects/MultiAlgoSelector.jsx
```diff
@@ -559,11 +609,32 @@ export default function MultiAlgoSelector({
...
+                                || (sel.target_repo_version ? { id: sel.target_repo_version, version: sel.target_repo_version } : null)
+                              const bridgeVersionHelperText = bridgeSourceUrl
+                                ? (
+                                  !bridgeVersionsLoaded
+                                    ? t('map_project.loading_versions', 'Loading versions...')
+                                    : hasSelectableBridgeVersions
+                                    ? t('map_project.bridge_version_description', 'Only versions with LLM support can be selected for bridge matching.')
+                                    : t('map_project.bridge_version_unavailable', 'No bridge source versions with LLM support are available.')
+                                )
+                                : t('map_project.bridge_version_source_required', 'Set a bridge source URL before choosing a version.')
```
paynejd: These four helper strings (`loading_versions`, `bridge_version_description`, `bridge_version_unavailable`, `bridge_version_source_required`) rely on inline `t(key, 'English default')` fallbacks and were not added to the `es`/`zh` translation files — whereas the other new keys in this PR (`repo_url`, `repo_version`, `not_vectorised`, `latest`) were added to all three locales. For consistency, add these four to en/es/zh as well so Spanish/Chinese users don't fall back to English here.

paynejd (commented):
Code review (automated assist by Claude, on behalf of Jonathan). Overall a clean, well-scoped implementation of #2559 — tests pass and lint is clean. A few inline notes below: one is a real (if low-probability) crash, the rest are correctness/UX/cleanup. Separately: the ticket's **migration of existing projects** to pin resolved `latest` is not in this PR (still unchecked in the issue) — so the original re-evaluation bug remains for projects that aren't re-saved. Worth confirming that's a tracked follow-up before closing #2559.

src/components/map-projects/MultiAlgoSelector.jsx
snyaggarwal: "latest" has a long running meaning in OCL -- latest released version. I dont think we should drift from that or redefine its meaning here.

If the "latest" is not eligible it will clearly say "Not Vectorised" -- that I feel is a clear and good enough indicator to clear the confusion

src/components/map-projects/projectTargetRepo.js
snyaggarwal: we rely on API ordering on many places in TBv2/3 in dropdowns. Also we dont want to implement ordering at multiple places (duplicate maintainence)

paynejd:
Reviewed the diff against #2559. The version-locking mechanics, helpers, and tests look solid and it's backward compatible. Two pieces of feedback — one design, one a small UX fix with code.

## 1. Resolution is config-time only — recommend resolving once at process start

Right now the "resolve `latest`/unspecified → concrete version" step only happens inside `MultiAlgoSelector` (i.e. when a core user opens the **Configuration** screen). At evaluation/run time, `MapProject` still:

- sends `bridge_repo.version` **only if** `target_repo_version` is already pinned (`buildProjectContext`), and
- falls back to the literal string `'latest'` for the mapped-sources fetch (`fetchMappedSources(bridgeUrl + bridgeVersion + '/')`, `bridgeVersion = ... || 'latest'`).

So a project that's never opened-and-saved in config (or created via API, or owned by a non-core user who can't see the selector) still floats on `latest` at runtime — which is exactly the failure mode this ticket is about. The same gap means `getAlgoLogExtras` won't record `repo_version` for those runs, so the new Discuss log won't show which version was actually used either. In other words, the bug is fixed for the *edit-and-save* path, not universally — consistent with the still-unchecked "migrate existing projects" item.

**Suggestion:** when a mapping/evaluation **process starts**, do a `$resolveReference` on the target repo URL to resolve it to a concrete, valid repo version, then **cache that resolved version for the lifetime of the process** and reuse it for every algo call in that run. We already do exactly this during automatch with ai-assistant prompt-template version resolution (resolve the default version once, pin it for the run). This guarantees that no matter what happens to the target content mid-process — a new version released, a version retired/un-vectorised — the *same* repo version is used start to finish. It also closes the gap above without depending on the user opening config, and gives a concrete value to log.

This is complementary to the config-time selector here (which is still the right place for an explicit user pin) — the process-start resolution is the safety net for everything that *isn't* explicitly pinned.

## 2. Warn the user when a pinned version is auto-reset (nit)

The auto-default effect silently overwrites a previously-saved `target_repo_version` when it's no longer in the repo's version list. That should surface as a warning so the user knows their pin changed. Wiring it through the existing `OperationsContext` alert channel:

```jsx
// imports
import { OperationsContext } from '../app/LayoutContext'

// inside MultiAlgoSelector(), near `const { t } = useTranslation()`
const { setAlert } = React.useContext(OperationsContext)
```

```jsx
// in the bridge-versions effect, where the default is applied:
const selectedVersion = loadedVersions.find(version => version?.id === sel?.target_repo_version)
if(!selectedVersion) {
  const defaultVersion = getDefaultBridgeRepoVersion(loadedVersions)
  if(defaultVersion?.id && defaultVersion.id !== sel?.target_repo_version) {
    const previousVersion = sel?.target_repo_version
    updateSelected(sel.__key, { target_repo_version: defaultVersion.id })
    if(previousVersion && setAlert)
      setAlert({
        message: t('map_project.bridge_version_reset', {version: defaultVersion.id, previous: previousVersion}),
        severity: 'warning'
      })
  }
}
```

Only fires when there *was* a non-empty saved version (initial resolution of an unset version stays silent), and it's single-fire — the next effect pass finds the new id in the list and short-circuits.

Translation keys (en/es/zh):

```json
"bridge_version_reset": "Bridge repo version changed to {{version}} because the previously selected version ({{previous}}) is no longer available."
"bridge_version_reset": "La versión del repositorio de Bridge se cambió a {{version}} porque la versión seleccionada anteriormente ({{previous}}) ya no está disponible."
"bridge_version_reset": "桥接仓库版本已更改为 {{version}}，因为先前选择的版本（{{previous}}）已不可用。"
```

(ESLint clean, JSON valid, existing `projectTargetRepo` tests unaffected.)

PR #45, opened 2026-06-05
Attribute debounced reranker calls to the auto-match run (ocl_online#115)

paynejd:
## Why

Live 9-row prod verification of the ocl_online#105 attribution chain (`automatch_run_id=1`, project 138 driven through the deployed Mapper) found one gap: **`reranker` was attributed `mapper-ui-manual`**, so only **1 of 9** rerank calls showed up under `?automatch_run_id=1`.

Root cause: `rerank()`'s `isBulk` flag did double duty — the `setAutoMatched` auto-propose side-effect **and** attribution. The explicit `processRerankWithConcurrency` call passes `isBulk=true` (attributed), but during a run candidates stream in asynchronously (scispacy cold-start especially), so the **debounced `scheduleRerank` path does most of the reranking** — and it called `rerank(rowIndex)` with no run flag → `mapper-ui-manual`, no `automatch_run_id`.

## What

- Decouple the two concerns: `rerank(_index, isBulk=false, isRunTraffic=isBulk)`. `isBulk` keeps driving `setAutoMatched`; `isRunTraffic` is attribution-only.
- `scheduleRerank` derives `isRunTraffic` from the **same** `isBulkMatchRunningRef && bulkMatchAlgoIds.length` condition it already gates bulk-vs-single behavior on, and passes it through the debounced call with `isBulk=false` (so the explicit bulk call's `setAutoMatched` isn't doubled).
- Adds reranker run/manual attribution contract tests.

## Verified
- `npm test` 167/167 · `npm run test:integration` 6/6 · eslint clean.
- The rest of the chain verified green in the same run: AutomatchRun create→complete (9/9); semantic/bridge/scispacy `$match`, `$resolveReference` reads, and AI `$invoke` all `automatch` under run 1; 98 api_transactions + 9 prompt_executions queryable by `automatch_run_id=1`.

Bare-references ocl_online#115 — closes after this lands, redeploys, and a re-run confirms `reranker` 9/9.

### DIFF
--- a/src/components/map-projects/MapProject.jsx
+++ b/src/components/map-projects/MapProject.jsx
@@ -2993,7 +2993,13 @@ const MapProject = () => {
     return result.concept_key
   }
 
-  const rerank = async (_index, isBulk=false) => {
+  // isBulk drives the auto-propose side-effect (setAutoMatched) for the
+  // processRerankWithConcurrency path. isRunTraffic is attribution-only
+  // (ocl_online#105 Phase 5) and defaults to isBulk; the debounced
+  // scheduleRerank path passes isRunTraffic=true WITHOUT isBulk so a bulk-run
+  // rerank is attributed to the run without doubling the setAutoMatched side
+  // effect the explicit bulk call already performs.
+  const rerank = async (_index, isBulk=false, isRunTraffic=isBulk) => {
     const index = isNumber(_index) ? _index : rowIndex
     if(!isNumber(index)) return null
     if(inFlightRerankRef.current.has(index)) {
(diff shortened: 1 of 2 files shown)


paynejd:
## Independent review — verdict: low risk, safe to merge & deploy

Reviewed + verified ahead of a merge-without-second-review.

**Root cause confirmed against prod analytics** (`api_transactions`, `algorithm_id='reranker'`): exactly the 8/9 vs 1/9 split described — 8 reranks `mapper-ui-manual` (no run_id), 1 `automatch` run 1, all project 138 in a single 67s window. Rows 0–7 went through the debounced `scheduleRerank` path (the gap); only the final row hit the explicit `processRerankWithConcurrency` sweep.

**Fix is surgical:** splits `rerank()`'s overloaded `isBulk` into `isBulk` (drives `setAutoMatched`) + `isRunTraffic` (attribution-only). `isBulk` is unchanged at both call sites, so the user-facing auto-propose behavior is provably untouched — change is purely additive on the attribution axis. Debounced path derives `isRunTraffic` from the same condition it already uses for bulk-vs-single gating (internally consistent).

**Checks:** `npm test` 167/167 · integration 6/6 · eslint clean · production webpack build compiles. `automatchRunRef` lifecycle traced — cleared after the awaited rerank sweep, so debounced reranks read the correct run_id.

**Residuals (minor):** attribution now matches the pre-existing gating decision, so no new misclassification is introduced; worst case is a one-off analytics miscount, never a functional/data-integrity issue. Note the new tests pin the header-builder contract, not the MapProject wiring — wiring confidence rests on code-read + the prod-data match + the planned post-deploy re-run.

— Claude Code review

PR #46, opened 2026-06-11
OpenConceptLab/ocl_issues#2560 | removed hardcoded fallbacks for CIEL bridge URL

### LINKED ISSUE #2560: Cleanup CIEL Bridge flow hardcoded fallbacks
CIEL bridge flow still hardcodes `/orgs/CIEL/sources/CIEL/` in multiple runtime and migration paths, and mapped-source loading still pins bridge discovery to `latest/` instead of an explicit persisted bridge version.

Here’s the same summary:

- Hardcoded CIEL bridge URL in the bridge algo editor: [MultiAlgoSelector.jsx#L565](https://github.com/OpenConceptLab/oclmap/blob/main/src/components/map-projects/MultiAlgoSelector.jsx#L565) uses `sel.target_repo_url ?? algo.target_repo_url ?? '/orgs/CIEL/sources/CIEL/'`, with the same value shown as the placeholder at [MultiAlgoSelector.jsx#L567](https://github.com/OpenConceptLab/oclmap/blob/main/src/components/map-projects/MultiAlgoSelector.jsx#L567).

- Runtime normalization fallback to CIEL when bridge URL is missing: [MapProject.jsx#L474-L490](https://github.com/OpenConceptLab/oclmap/blob/main/src/components/map-projects/MapProject.jsx#L474-L490) builds `projectContext.bridge_repo` and falls back to `'/orgs/CIEL/sources/CIEL/'` for `ocl-ciel-bridge`.

- Hardcoded `latest` bridge version during mapped-source preload: [MapProject.jsx#L513-L517](https://github.com/OpenConceptLab/oclmap/blob/main/src/components/map-projects/MapProject.jsx#L513-L517) loads the bridge source from `bridgeAlgoFromApi.target_repo_url || '/orgs/CIEL/sources/CIEL/'` and then always fetches `bridgeUrl + 'latest/'`.

- Same CIEL URL fallback used when rendering `BridgeMatch`: [MapProject.jsx#L4203-L4215](https://github.com/OpenConceptLab/oclmap/blob/main/src/components/map-projects/MapProject.jsx#L4203-L4215) uses `bridgeAlgo?.target_repo_url || '/orgs/CIEL/sources/CIEL/'`.

- Bridge version exists in the normalization model, but isn’t populated in this flow: [normalizers.js#L111-L113](https://github.com/OpenConceptLab/oclmap/blob/main/src/components/map-projects/normalizers.js#L111-L113) reads `projectContext.bridge_repo.version`, while save validation only enforces target repo version at [MapProject.jsx#L1177-L1188](https://github.com/OpenConceptLab/oclmap/blob/main/src/components/map-projects/MapProject.jsx#L1177-L1188). (related ticket to this https://github.com/OpenConceptLab/ocl_issues/issues/2559)

- Legacy migration scripts also backfill missing bridge metadata with the same hardcoded CIEL path: [dump-projects.sql#L63-L67](https://github.com/OpenConceptLab/oclmap/blob/main/scripts/migrate-candidates-to-v2/dump-projects.sql#L63-L67) and [fix-algorithm-canonicals.sql#L45-L48](https://github.com/OpenConceptLab/oclmap/blob/main/scripts/migrate-candidates-to-v2/fix-algorithm-canonicals.sql#L45-L48).

**Merge https://github.com/OpenConceptLab/ocl_issues/issues/2559 PRs before this**

snyaggarwal:
## Linked Issue

Refs OpenConceptLab/ocl_issues#2560

Sibling: https://github.com/OpenConceptLab/react-bridge-match/pull/4

### DIFF
--- a/src/components/map-projects/MapProject.jsx
+++ b/src/components/map-projects/MapProject.jsx
@@ -513,15 +513,6 @@ const MapProject = () => {
         const response = await service.get();
         const _algos = response?.data?.results || []
         setApiAlgos(_algos);
-        const bridgeAlgoFromApi = find(_algos, a => ['ocl-bridge', 'ocl-ciel-bridge'].includes(a.type))
-        if(bridgeAlgoFromApi?.target_repo_url) {
-          const bridgeUrl = bridgeAlgoFromApi.target_repo_url
-          const bridgeVersion = bridgeAlgoFromApi.target_repo_version || ''
-          if(!bridgeVersion)
-            return
-          fetchMappedSources(bridgeUrl + bridgeVersion + '/', sources =>
-            setBridgeMappedSources(prev => ({...prev, [bridgeUrl]: sources})))
-        }
       } catch {
         // pass
       }
--- a/src/components/map-projects/MapProject.jsx
+++ b/src/components/map-projects/MapProject.jsx
@@ -483,14 +483,8 @@ const MapProject = () => {
     // URL captured on the algo's bridge_repo metadata (set via the
     // MultiAlgoSelector bridge canonical_url field, PR2b); fall back to the
     // derived form (https://ns.openconceptlab.org + relative_url) when only
-    // the relative URL is known. If target_repo_url is missing entirely —
-    // the algo was added without ever editing the bridge source URL field —
-    // fall back to the type's well-known default (matches the placeholder
-    // shown in MultiAlgoSelector). Without this fallback, normalization
-    // silently produces zero candidates for bridge-only flows.
-    const BRIDGE_DEFAULT_RELATIVE_URL = { 'ocl-ciel-bridge': '/orgs/CIEL/sources/CIEL/' }
+    // the relative URL is known.
     const bridgeRelativeUrl = bridgeAlgo?.target_repo_url
-      || BRIDGE_DEFAULT_RELATIVE_URL[bridgeAlgo?.type]
     if(bridgeAlgo && bridgeRelativeUrl) {
       const explicitBridgeCanonical = bridgeAlgo?.bridge_repo?.canonical_url
       ctx.bridge_repo = {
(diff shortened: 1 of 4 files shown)


src/components/map-projects/MapProject.jsx
```diff
@@ -520,9 +514,11 @@ const MapProject = () => {
...
         setApiAlgos(_algos);
         const bridgeAlgoFromApi = find(_algos, a => ['ocl-bridge', 'ocl-ciel-bridge'].includes(a.type))
-        if(bridgeAlgoFromApi) {
-          const bridgeUrl = bridgeAlgoFromApi.target_repo_url || '/orgs/CIEL/sources/CIEL/'
-          const bridgeVersion = bridgeAlgoFromApi.target_repo_version || 'latest'
+        if(bridgeAlgoFromApi?.target_repo_url) {
+          const bridgeUrl = bridgeAlgoFromApi.target_repo_url
+          const bridgeVersion = bridgeAlgoFromApi.target_repo_version || ''
+          if(!bridgeVersion)
+            return
```
paynejd: **This preload is a no-op against current production data.**

I queried prod `/match-algorithms/`: the `ocl-ciel-bridge` algo has `target_repo_url: "/orgs/CIEL/sources/CIEL/"` but **no `target_repo_version`**. So `bridgeVersion` resolves to `''` and this block hits `if(!bridgeVersion) return` on every load — `fetchMappedSources` here never runs in prod (before this PR it fetched `…/CIEL/latest/`).

The bridge flow still works because the resolve-effect below (~L587-617) populates the same `bridgeMappedSources` state via `/$resolveReference/` and backfills the version into `algosSelected` — and that's the only writer the render at ~L4358 actually depends on.

Net: as written, this effect is dead code for the only bridge algo that exists in prod. Two clean options:
- **Remove this effect** as redundant with the resolve-effect, or
- **Add `target_repo_version` to the algo definition** on the backend (this is #2559's "Add target_repo_version in bridge algo", which is checked off in the issue but isn't present in prod).

Not a blocker — approving — but worth resolving so the code doesn't read as if the preload does something it never does in prod.

paynejd (approved):
Approving — clean, surgical removal of the CIEL/`latest` fallbacks. The oclmap↔react-bridge-match `mappedRepoURLs` wiring is consistent (gated on non-empty `mappedSrcs`), and all 174 unit tests pass on the branch.

One non-blocking note inline on the API preload effect (finding #1). Two minor follow-ups for later: the migration SQL scripts named in #2560 (item 6) are intentionally left untouched — worth a one-line note on the issue so the remaining scope is tracked; and the sibling react-bridge-match PR still keeps a `bridgeRepoVersion || 'latest'` fallback (unreachable in this flow, but contradicts the cleanup intent).

PR #47, opened 2026-06-11
OpenConceptLab/ocl_online#117 | fixing CSV download for bridge candidates

snyaggarwal:
## Linked Issue

Refs: OpenConceptLab/ocl_online#117

### DIFF
--- a/src/components/map-projects/MapProject.jsx
+++ b/src/components/map-projects/MapProject.jsx
@@ -101,7 +101,7 @@ import { DEFAULT_ENCODER_MODEL } from './rerankerModels'
 import { normalizeAlgorithmInvocation, lookupStatusRank, buildRecommendableConceptEntry, stripConstantClassAndDatatype, buildLookupConceptUrl } from './normalizers'
 import { parseConceptKey } from './conceptKey'
 import { getDefaultTargetRepoVersion, getProjectTargetRepoVersion, getTargetRepoVersionFromUrl, getTargetRepoVersionId } from './projectTargetRepo'
-import { buildQualityRowViews, conceptBelongsToTargetRepo, conceptForMapping, resolveAICandidateID } from './viewBuilders.js'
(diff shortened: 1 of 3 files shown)


paynejd (approved):
**Approving** — faithful to the ticket and cleanly factored. Nice call putting the whole fix in OSS `viewBuilders.js` rather than the premium `@snyaggarwal/bridge-match` module: the CSV is now self-sufficient. De-dup/sort/convergence and the map_type regex-extraction all check out, and both new tests pass locally.

Two small changes to make:

**1. Drop the `targetRawScore` fallback.**
In `buildBridgeTargetDownloadEntries`:
```js
targetRawScore: toNumberOrNull(qualityView?.candidate?.score) ?? bridgeRawScore,
```
Remove the `?? bridgeRawScore`. A bridge-only target has no independent raw match score, so falling back to the bridge's raw just duplicates "Bridge Raw Score" and reads as a target match it isn't. Let it be null — `formatBridgeTargetDownloadEntry` already `compact()`s out null score segments, so the field then appears only when the target also surfaced via a direct algorithm (a real, distinct number).

**2. Make the score labels explicit about bridge vs target.**
In `formatBridgeTargetDownloadEntry`, each score should say what it is:
- `Bridge Raw Score` ✓ keep
- `Unified Score` → **`Target Unified Score`** (the target's reranked/unified score)
- `Raw Score` → **`Target Raw Score`**

So a cell reads `Bridge Raw Score: 0.95 | Target Unified Score: 98.6 | Target Raw Score: 0.91`, and a bridge-only target reads `Bridge Raw Score: 0.95 | Target Unified Score: 98.6`. The `formatBridgeTargetDownloadEntry` test assertions will need the new labels too.

**Non-blocking:** the PR `Refs:` #117 rather than `Closes:`, and cross-repo (oclmap → ocl_online) won't auto-close on GitHub — so OpenConceptLab/ocl_online#117 will need a manual close when this merges.

PR #48, opened 2026-06-19
OpenConceptLab/ocl_issues#2580 | Stopping auto-retrieving of candidates on row click

### LINKED ISSUE #2580: OCL Mapper: Auto-match row control — rename "Refresh" to "Retrieve/Refresh Candidates" and stop auto-running on row-click
Part of #2316.

Stop auto-retrieving candidates **when a row is clicked**; make retrieval an explicit button action.

### Current
Clicking a row calls `fetchAllCandidatesForRow(...)` (`MapProject.jsx` ~L2424–2433). Manual control = `RefreshIcon` in `Candidates.jsx` (`onRefreshClick`, tooltip `map_project.refresh_candidates_tooltip`).

### Acceptance criteria
- [x] Row click selects + opens the right panel **without** triggering retrieval.
- [x] Rename Refresh → **"Retrieve/Refresh Candidates"** (label + i18n key); retrieval only on button click.
- [x] **Decide & document:** retrieve-only, full row auto-match, or both?

Components: `Candidates.jsx`, `MapProject.jsx`.

snyaggarwal:
## Linked Issue

Closes OpenConceptLab/ocl_issues#2580

### DIFF
--- a/src/components/map-projects/Candidates.jsx
+++ b/src/components/map-projects/Candidates.jsx
@@ -657,7 +657,7 @@ const Candidates = ({rowIndex, alert, setAlert, rowState, conceptCache, targetCa
           }
           <Tooltip title={t('map_project.refresh_candidates_tooltip')}>
             <span>
-              <IconButton color='secondary' onClick={onRefreshClick} disabled={!areAlgoRun} size='small' sx={{margin: '0 4px'}}>
+              <IconButton color='secondary' onClick={onRefreshClick} disabled={rowStage !== undefined && !areAlgoRun} size='small' sx={{margin: '0 4px'}}>
                 <RefreshIcon />
               </IconButton>
             </span>
--- a/src/components/map-projects/Candidates.jsx
+++ b/src/components/map-projects/Candidates.jsx
@@ -657,7 +657,7 @@ const Candidates = ({rowIndex, alert, setAlert, rowState, conceptCache, targetCa
           }
           <Tooltip title={t('map_project.refresh_candidates_tooltip')}>
             <span>
-              <IconButton color='secondary' onClick={onRefreshClick} disabled={rowStage !== undefined && !areAlgoRun} size='small' sx={{margin: '0 4px'}}>
+              <IconButton color='secondary' onClick={onRefreshClick} disabled={algoStagesValue.some(v => v === 0)} size='small' sx={{margin: '0 4px'}}>
                 <RefreshIcon />
               </IconButton>
             </span>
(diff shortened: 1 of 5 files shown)


src/components/map-projects/Candidates.jsx
```diff
@@ -657,7 +657,7 @@ const Candidates = ({rowIndex, alert, setAlert, rowState, conceptCache, targetCa
           }
           <Tooltip title={t('map_project.refresh_candidates_tooltip')}>
             <span>
-              <IconButton color='secondary' onClick={onRefreshClick} disabled={!areAlgoRun} size='small' sx={{margin: '0 4px'}}>
+              <IconButton color='secondary' onClick={onRefreshClick} disabled={rowStage !== undefined && !areAlgoRun} size='small' sx={{margin: '0 4px'}}>
```
paynejd: **Failed rows can't retry.** The new condition correctly re-enables a fresh row (`rowStage === undefined`), but with stage values `0 = in-flight`, `1 = done`, `-2 = failed`, `areAlgoRun` is `false` whenever **any** algo failed or is partial — so a row whose retrieval errors (`-2`) stays disabled. Previously that row could be re-clicked to re-fetch; now that auto-retrieve-on-click is gone, this button is the only path, so the row gets stuck. Gate on *in-flight* instead — enables fresh + failed + completed, disables only during an active fetch:

```suggestion
              <IconButton color='secondary' onClick={onRefreshClick} disabled={algoStagesValue.some(v => v === 0)} size='small' sx={{margin: '0 4px'}}>
```

(`algoStagesValue` already excludes `recommend`, so this preserves the prior recommend behavior.)

src/components/map-projects/MapProject.jsx
```diff
@@ -2468,10 +2468,8 @@ const MapProject = () => {
       if(!appliedFacets[csvRow.__index] && !isEmpty(_filters) && _filters) {
         setAppliedFacets({...appliedFacets, [csvRow.__index]: getAppliedFacetFromQueryParam(_filters)})
       }
-      fetchAllCandidatesForRow(null, csvRow, 0, undefined, undefined, getAppliedFacetFromQueryParam(_filters))
       if(isEmpty(getFacetsForRow(csvRow.__index)))
         getFacets(true, csvRow.__index)
```
paynejd, line 2472: **Non-blocking:** `getFacets` still fires on row-open. It's a lighter call than candidate retrieval (no algorithms run) and it readies the filter panel, so keeping it is defensible — but if the intent is *zero* auto-fetch on click, consider deferring this to the first Retrieve as well. Fine to leave as-is for this ticket.

paynejd (approved):
### Review — stop auto-retrieving candidates on row click

Approving. ✅ Small, correct, and on-target for #2580: the row-click `fetchAllCandidatesForRow` is removed so opening a row no longer retrieves, the Refresh control is renamed to “Retrieve/Refresh Candidates” across all three locales, and the AC's open question is resolved in the ticket (retrieve-only now; retrieve+automatch as a separate action later). Confirmed stage semantics: `0` = in-flight, `1` = done, `-2` = failed.

One item worth fixing before merge (inline suggestion): the new `disabled` condition re-enables fresh rows but leaves **failed / partially-failed** rows unable to retry — and that path is now only reachable via this button. Plus a non-blocking note on facets still loading on row-open.

PR #50, opened 2026-06-21
OpenConceptLab/ocl_issues#2579 | AutoMatch can run on selected rows

### LINKED ISSUE #2579: OCL Mapper: Auto-match scope — add "Selected rows" option
Part of #2316.

Add a **"Selected rows"** scope to the Auto Match dialog, alongside the existing **Unmapped only** / **All rows**, so auto-match can run against just the rows multi-selected on the left panel (#2342).

### Acceptance criteria
- [x] Third scope option "Selected rows" in `AutoMatchDialog.jsx`'s `RadioGroup` (today `autoMatchUnmappedOnly` boolean → make a 3-value scope `unmapped | all | selected`).
- [x] Enabled only with an active selection; shows selected count, disabled with helper text otherwise.
- [x] "X out of Y" summary + Retrieve-candidates / AI-analysis options respect the chosen scope.

Depends on #2342. Components: `AutoMatchDialog.jsx`, `MapProject.jsx`.

snyaggarwal:
## Linked Issue

Closes OpenConceptLab/ocl_issues#2579


src/components/map-projects/MapProject.jsx
```diff
@@ -2677,7 +2704,10 @@ const MapProject = () => {
     })
   }
 
-  const getSelectedRowIndexes = () => selectedRowIds.map(id => parseInt(id)).filter(Number.isFinite)
+  const getSelectedRowIndexes = (_rows = data) => {
+    const selectedIds = new Set(selectedRowIds.map(id => id?.toString()))
+    return _rows.filter(_row => selectedIds.has(_row.__index?.toString())).map(_row => _row.__index)
+  }
```
paynejd, line 2703: **Robustness regression:** `data` initializes to `false` (L157), so when this runs with no args (e.g. `onGetCandidates` before the sheet finishes parsing — the button is gated on `file`, not `data`), `_rows` is `false` and `.filter` throws. The previous `selectedRowIds.map(parseInt)` version never touched `data`. Cheap guard (`isArray` is already imported):

```suggestion
  const getSelectedRowIndexes = (_rows = data) => {
    if(!isArray(_rows)) return []
    const selectedIds = new Set(selectedRowIds.map(id => id?.toString()))
    return _rows.filter(_row => selectedIds.has(_row.__index?.toString())).map(_row => _row.__index)
  }
```

src/components/map-projects/MapProject.jsx
```diff
@@ -1591,6 +1591,16 @@ const MapProject = () => {
...
     abortRef.current = false;
+    const selectedRowIndexes = getSelectedRowIndexes(rows)
+    const selectedRowIndexSet = new Set(selectedRowIndexes)
+    const isAutoMatchUnmappedOnly = autoMatchScope === 'unmapped'
+    const isAutoMatchAllRows = autoMatchScope === 'all'
+    const isAutoMatchSelectedRows = autoMatchScope === 'selected'
+    const selectedRowsLogExtras = isAutoMatchSelectedRows ? {
+      selected_rows_count: selectedRowIndexes.length,
+      row_indexes: selectedRowIndexes,
+      selected_row_indexes: selectedRowIndexes
```
paynejd, line 1602: `row_indexes` and `selected_row_indexes` are the same array, and this object is spread into **both** `auto_match_started` and `auto_match_finished` — which `projectLog` POSTs to `…/logs/` (L1337). A large selection therefore persists the same index array up to 4× per run, while only `selected_rows_count` is read (in `ProjectLogs`). Drop the duplicate (keep whichever the analytics pipeline expects; if neither, `selected_row_indexes` is clearest) — and worth deciding whether the full array is needed at all vs. just the count:

```suggestion
      selected_rows_count: selectedRowIndexes.length,
      selected_row_indexes: selectedRowIndexes
```

src/components/map-projects/AutoMatchDialog.jsx
```diff
@@ -98,22 +119,30 @@ const AutoMatchDialog = ({
...
             aria-labelledby="automatch-rows"
             name="automatch-rows"
-            onChange={() => setAutoMatchUnmappedOnly(!autoMatchUnmappedOnly)}
+            value={autoMatchScope}
+            onChange={event => setAutoMatchScope(event.target.value)}
           >
+            <FormControlLabel
+              value="all"
+              control={<Radio />}
+              label={`${t('map_project.all_rows')} (${allRowsCount.toLocaleString()})`}
```
paynejd: **Relabel (per review discussion):** this option labeled "All Rows" actually targets Unmapped + Proposed (it skips Approved — see `MapProject.jsx` L1778), which collides with the main-page **All** bucket. Rename to **"Unmapped & Proposed"**. `map_project.all_rows` is used only here, so add a dedicated key:

```suggestion
              label={`${t('map_project.unmapped_and_proposed')} (${allRowsCount.toLocaleString()})`}
```

Then add the key to en/es/zh and remove the now-unused `all_rows` (please verify es/zh wording for consistency):
- en: `"unmapped_and_proposed": "Unmapped & Proposed"`
- es: `"unmapped_and_proposed": "No mapeadas y propuestas"`
- zh: `"unmapped_and_proposed": "未映射和已提议"`

A true "All (incl. Approved)" / start-from-scratch scope (with a warning) is tracked separately in #2589.

paynejd (commented):
Reviewed against #2579. Solid, well-scoped feature — the boolean→3-value `autoMatchScope` migration is clean and complete (no dangling refs), the `getSelectedRowIndexes` refactor is actually *more* robust (stringified set comparison handles the string-id vs numeric-`__index` mismatch), and the selection-pruning effect keeps the dialog count and the processed set consistent. i18n parity across en/es/zh is good.

Four inline notes: a robustness regression (falsy `data`), a log-payload duplication, the agreed **"All Rows" → "Unmapped & Proposed"** relabel, and an unused i18n key.

**Design decision (from review):** the "All Rows" option silently excluded Approved rows and collided with the main-page **All** bucket. Resolution — relabel to "Unmapped & Proposed" here; a true "All (incl. Approved)" scope with a warning/confirm is tracked in #2589.

Non-blocking nits (left out of inline):
- `ProjectLogs` filters on the *formatted* string `'Selected Rows'`; filtering the raw `selected_rows` token before `formatSubAction` would be less brittle.
- "Selected (N)" can render larger than "Unmapped & Proposed (M)" when Approved rows are selected (Selected counts them; the other scope doesn't) — cosmetic.
- No unit tests for the scope→`rowsToProcess` branching; the repo has Jest under `__tests__/` if you'd like a couple.

src/components/map-projects/AutoMatchDialog.jsx
```diff
@@ -47,11 +48,31 @@ const AutoMatchDialog = ({
...
+    }
+  }, [autoMatchScope, hasUnmappedRows, setAutoMatchScope])
 
   const getHelperTextForAutoMatchUnmapped = () => {
-    if (autoMatchUnmappedOnly) {
+    if (autoMatchScope === 'selected') {
+      if (hasSelectedRows) {
+        return t('map_project.auto_match_selected_rows_note', {count: selectedRowCount.toLocaleString()});
+      }
+      return '';
```
paynejd: `auto_match_selected_rows_note_no_count` was added to all 3 locales but is never referenced — this branch returns `''`, and it's effectively unreachable anyway (the `selected` radio is `disabled` without a selection). Either wire it here or drop the key from en/es/zh.

PR #60, opened 2026-09-21
OpenConceptLab/ocl_issues#2773 | Mapper preview

snyaggarwal:
## Linked Issue

Refs: 
OpenConceptLab/ocl_issues#2773
OpenConceptLab/ocl_issues#2781


paynejd (commented):
Reviewed against ocl_issues#2773, #2762 and #2781, and cross-checked the contract against the backend branch.

## ⚠️ Dependency: this PR cannot merge before OpenConceptLab/oclapi2#900

**oclapi2#900 (`mapper-preview`) is a hard dependency of this PR, and it is not mergeable yet** — my review there on 2026-09-21 lists four blockers (500 on project re-create after delete, self-serve quota refund via negative `units`, inverted kill switch, hardcoded `pk=16` in the group migration). Every capability name, error code and meter this PR reads is defined there. Merging this first ships a UI that reads fields the API does not yet return.

Two of those backend items are **shared contract decisions that have to move in both PRs together** — they are not "backend-only":

1. **What a match operation is.** The backend meters `$match` as `units=len(rows)`; #2762/#2782 specify **row–algorithm pairs**. This PR's estimate ([`AutoMatchDialog.jsx#L88`](https://github.com/OpenConceptLab/oclmap/blob/657aa52/src/components/map-projects/AutoMatchDialog.jsx#L88)) and pre-truncation ([`MapProject.jsx#L2091-L2094`](https://github.com/OpenConceptLab/oclmap/blob/657aa52/src/components/map-projects/MapProject.jsx#L2091-L2094)) both multiply by `algorithmCount`, i.e. they implement the **spec**, not the backend. As it stands a 4-algorithm run is estimated at 4x what the server actually charges — the UI will truncate runs the server would have allowed. Whichever definition wins, both sides change.
2. **`0 = unlimited`.** #2782's runbook says "set caps to 0" to kill the feature; in code `0` means unlimited, so following the runbook uncaps everyone. This PR inherits the same ambiguity in its own form — see the `autoMatchRows.js` comment below, where a *missing* capability is read as unlimited by one helper and as blocked by another.

Please hold this PR until #900's blockers land, then re-point it at the final contract.

## The frontend review

The architecture is good: `autoMatchRows.js` is a clean extraction with real tests, `MapperQuotaChip` / `PreviewLimitDialog` are well factored, and all three locales (en/es/zh) are complete and consistent — that last one is genuinely rare and appreciated.

The problems cluster in one place, and it's the same seam as the backend: **the UI and the quota disagree about what has been spent and what is left.** Four blockers, four should-fixes, all inline.

**Blockers**
1. Bulk matching swallows a preview-limit 403 as a successful empty result — the exact silent failure #2773 asked us to fix.
2. A limit error on a plain row click leaves the row spinning forever.
3. The quota cache is never refreshed on the ordinary row-click path, which is what makes stale counts (and #1) reachable.
4. Existing sessions never see the feature at all — `permissions` is dropped on refresh, so `hasCapability` stays false until re-login. Combined with #900's missing backfill for the 14 `mapper-approved` users, deploy day is rough for every current Mapper user.

Not approving yet. Happy to pair on the metering definition — as above, I don't think it should be settled in review comments on two separate PRs.

src/components/map-projects/MapProject.jsx
```diff
@@ -2044,7 +2084,22 @@ const MapProject = () => {
       }))
 
     setTimeout(async () => {
-      const rowsToProcess = getRowsToProcess(rows, rowStatuses, autoMatchScope, selectedRowIndexes)
+      const preview = getMapperPreview()
+      const previewEligibleRowIndexes = getPreviewEligibleRowIndexes(rows, preview)
+      let rowsToProcess = getRowsToProcess(rows, rowStatuses, autoMatchScope, selectedRowIndexes, previewEligibleRowIndexes)
```
paynejd, line 2111: **Blocker — bulk matching treats a preview-limit 403 as a successful empty match.**

This is the bulk entry point, but the fix for #2773 only landed on the per-row path. `processBatch` (~[L1942-L1965](https://github.com/OpenConceptLab/oclmap/blob/657aa52/src/components/map-projects/MapProject.jsx#L1942-L1965)) calls `service.post(payload, token, headers, query)` with `raw` defaulting to `false`, and `APIService.sendRequest` resolves rather than throws in that mode:

```js
// src/services/APIService.js:145
return error.response ? error.response.data : error.message;
```

So a 403 `{error_code: 'mapper_match_operations_exceeded', detail: ...}` lands in the `try`, not the `catch`. The result:

```js
forEach(rowBatch, __row => {
  markAlgo(__row.__index, algo.id, 1)              // marked SUCCEEDED
  log({action: 'algo_finished', ...}, __row.__index) // logged as finished
})
return response.data || [];                         // error body has no .data -> []
```

Every row in the batch is marked `algo_finished` with zero candidates, no alert, no `PreviewLimitDialog`. The user sees a completed run that found nothing — indistinguishable from a genuine no-match, which is precisely the failure mode #2773 exists to eliminate.

`processBatch` needs the same `isPreviewLimitError(response)` check the per-row path got, before the success `forEach`:

```js
const response = await service.post(...);
if(isPreviewLimitError(response)) {
  forEach(rowBatch, __row => markAlgo(__row.__index, algo.id, -2))
  handlePreviewLimitError(response)
  throw new PreviewLimitAbort()   // or similar, to stop the remaining queue
  }
```

Worth deciding explicitly whether hitting the limit mid-run should **abort the remaining queue** or let it drain — every subsequent batch will 403 too, so draining means N pointless round trips and N more chances to mismark rows. I'd abort.

src/components/map-projects/MapProject.jsx
```diff
@@ -3162,6 +3239,40 @@ const MapProject = () => {
...
+
+  const handlePreviewLimitError = (response, rowId) => {
+    if(!isPreviewLimitError(response))
+      return false
+    setAlert({message: response?.detail || t('unknown_error'), severity: 'error'})
+    setIsLoadingInDecisionView(false)
+    if(isNumber(rowId) && restoreRefreshRowStage(rowId))
+      refreshMapperQuotaCache()
+    return true
+  }
```
paynejd: **Blocker — a limit error on a plain row click leaves the row spinning forever.**

`handlePreviewLimitError` returns `true` at L3274, and every caller treats `true` as "handled, stop here":

```js
if(handlePreviewLimitError(response, __row.__index))
  return                       // L3390 - skips markAlgo(..., -2) on L3394
```

But the only thing that resets the row's stage inside this function is `restoreRefreshRowStage(rowId)`, and that returns `false` unless a snapshot exists — snapshots are written **only** by `onRefreshClick` ([L2856](https://github.com/OpenConceptLab/oclmap/blob/657aa52/src/components/map-projects/MapProject.jsx#L2856)). So on the common path (user clicks a row, no Refresh button involved) the row stays at stage `0` — in-flight — with a spinner that never resolves. The alert fires, but the row is stuck until reload.

The same `refreshMapperQuotaCache()` on L3273 is gated behind that snapshot too, so the row that just burned an operation also doesn't refresh the meter (see my comment on L3455).

Suggested shape — pass the algo through so the stage can always be marked failed:

```suggestion
  const handlePreviewLimitError = (response, rowId, algoId) => {
    if(!isPreviewLimitError(response))
      return false
    setAlert({message: response?.detail || t('unknown_error'), severity: 'error'})
    setIsLoadingInDecisionView(false)
    if(isNumber(rowId)) {
      if(!restoreRefreshRowStage(rowId) && algoId)
        markAlgo(rowId, algoId, -2)
      refreshMapperQuotaCache()
    }
    return true
  }
```

That needs the three call sites updated to pass the id — [L3390](https://github.com/OpenConceptLab/oclmap/blob/657aa52/src/components/map-projects/MapProject.jsx#L3390) (`algoId`), [L3568](https://github.com/OpenConceptLab/oclmap/blob/657aa52/src/components/map-projects/MapProject.jsx#L3568) (`'ocl-scispacy-loinc'`), [L4008](https://github.com/OpenConceptLab/oclmap/blob/657aa52/src/components/map-projects/MapProject.jsx#L4008) (`bridgeAlgoId`) — all three have it to hand. Applying the suggestion alone is safe (it just no-ops the `markAlgo`), so please do both.

src/components/map-projects/MapProject.jsx
```diff
@@ -3337,6 +3452,8 @@ const MapProject = () => {
           markAlgo(__row.__index, nextAlgo.id, 0)
           fetchAllCandidatesForRow(nextAlgo.id, __row, offset, _retired, scrollToBottom, _filters, forceReload)
         } else {
+          if(clearRefreshRowStageSnapshot(__row.__index))
+            refreshMapperQuotaCache()
```
paynejd: **Blocker — the quota cache is never refreshed on the ordinary row-click path.**

This is the **success** path of per-row matching, and the refresh is gated on `clearRefreshRowStageSnapshot(__row.__index)`. That returns `true` only when a snapshot exists, and snapshots are written in exactly one place:

```js
// L2855-L2859 - onRefreshClick, and nowhere else
const onRefreshClick = () => {
  refreshRowStageSnapshotRef.current[rowIndex] = ...
```

So the meter refreshes after the **Refresh button**, and never after a normal row click — even though both consume the same match operations. A user working down the grid burns their quota against a `MapperQuotaChip` that keeps showing the count from page load, until something else happens to refresh it.

That stale read is also what makes the other two findings reachable: `getMapperPreview()` at [L2087](https://github.com/OpenConceptLab/oclmap/blob/657aa52/src/components/map-projects/MapProject.jsx#L2087) sizes the bulk pre-truncation off the same cache, so the truncation is computed from a quota that may be long spent — and then the server 403s mid-run into the swallow-bug above.

The snapshot bookkeeping and the meter refresh are independent concerns; they shouldn't share a condition:

```suggestion
          clearRefreshRowStageSnapshot(__row.__index)
          refreshMapperQuotaCache()
```

Same gating at [L3392](https://github.com/OpenConceptLab/oclmap/blob/657aa52/src/components/map-projects/MapProject.jsx#L3392), [L3570](https://github.com/OpenConceptLab/oclmap/blob/657aa52/src/components/map-projects/MapProject.jsx#L3570) and [L4012](https://github.com/OpenConceptLab/oclmap/blob/657aa52/src/components/map-projects/MapProject.jsx#L4012) — all four want the same treatment.

If the concern is request volume, debounce `refreshMapperQuotaCache` rather than skipping it; a stale quota is worse than an extra `GET /user/`.

src/common/utils.js
```diff
@@ -264,14 +264,29 @@ export const arrayToCSV = objArray => {
...
 
+export const refreshCurrentUserCapabilitiesCache = callback => {
+  APIService.user().get(null, null, {includeCapabilities: true}).then(response => {
+    if(response.status === 200) {
+      const currentUser = getCurrentUser()
+      if(currentUser) {
+        localStorage.setItem('user', JSON.stringify({
+          ...currentUser,
+          capabilities: response.data?.capabilities || []
+        }));
```
paynejd: **Blocker — existing sessions never gain `permissions`, so nobody with a live session sees the feature.**

This merge keeps `capabilities` and drops `permissions`. But every gate in the app reads `permissions`, not `capabilities`:

```js
// utils.js:1140
export const hasCapability = (user, capability) =>
  Boolean((user || getCurrentUser())?.permissions?.includes(capability))
```

`getMapperPreview()` uses it for `hasAccess`, `hasAIAssistant`, `hasCustomAlgorithms` and `hasOrgProjects` (L1153-L1156), and `MapProject.jsx:527` gates the AI Assistant UI on it. So for any user whose cached `user` object predates this release — i.e. everyone with a live session on deploy day — `permissions` stays `undefined` forever, `hasCapability` returns `false`, and the AI Assistant and quota chip simply don't appear. Calling this function makes it *look* fixed, which is worse than not calling it.

The backend already returns both under the same query param (`core/users/serializers.py:219-221` pops `permissions` and `capabilities` together on `includeCapabilities`), so this is a one-line merge:

```suggestion
        localStorage.setItem('user', JSON.stringify({
          ...currentUser,
          capabilities: response.data?.capabilities || [],
          permissions: response.data?.permissions || currentUser.permissions || []
        }));
```

Note this interacts with oclapi2#900: there is no backfill for the 14 existing `mapper-approved` users, so on deploy they lose the underlying permission *and* the client can't re-read it. Both halves need to be right or current Mapper users are locked out.

src/components/map-projects/autoMatchRows.js
```diff
@@ -1,19 +1,30 @@
-export const getRowsToProcess = (rows, rowStatuses, autoMatchScope, selectedRowIndexes = []) => {
+export const getPreviewEligibleRowIndexes = (rows, preview) => {
+  if(!Array.isArray(rows))
+    return []
+  const rowsPerProject = preview?.rowsPerProject || {}
+  if(rowsPerProject.unlimited || rowsPerProject.limit === null || rowsPerProject.limit === undefined)
+    return null
```
paynejd: **A missing capability is read as *unlimited* here and as *blocked* everywhere else.**

`limit === null || limit === undefined` returning `null` means "no restriction, every row eligible". But `toMeter` in `utils.js` reads the identical state as zero:

```js
// utils.js:1147-1150
const limit = entry.limit === undefined ? null : entry.limit
const unlimited = limit === 0                    // null is NOT unlimited
return {limit, used, unlimited, remaining: unlimited ? null : Math.max((limit || 0) - used, 0)}
//                                                           null -> 0 - 0 -> 0 remaining
```

So when a capability is absent from the payload, `AutoMatchDialog` computes `rowsRemaining = 0` → `isPreviewQuotaExhausted` → "No more rows are left in your preview quota" with Submit disabled, while this helper returns `null` and the grid happily enables every row for individual matching. Two surfaces, opposite conclusions, same data.

The backend docstring is explicit that `None` means blocked, not unlimited — `0` is the unlimited sentinel. Only `unlimited` should open the gate:

```suggestion
  if(rowsPerProject.unlimited)
    return null
  if(rowsPerProject.limit === null || rowsPerProject.limit === undefined)
    return []
```

Worth adding a case to `__tests__/autoMatchRows.test.js` pinning it — the current suite covers `unlimited: true` and a numeric limit, but not the missing-capability shape, which is the one that will actually show up if a capability name is ever misspelled or a group is misconfigured.

(This is the client-side face of the `0 = unlimited` question on oclapi2#900 — if that sentinel changes there, this changes too.)

src/components/map-projects/AutoMatchDialog.jsx
```diff
@@ -45,25 +47,55 @@ const AutoMatchDialog = ({
...
+  const algorithmCount = Math.max(algosSelected.length, 1)
+  const estimatedOperations = rowsToMatchCount * algorithmCount
+  const operationsRemaining = preview.matchOperations.unlimited ? null : preview.matchOperations.remaining
+  const rowsRemaining = preview.rowsPerProject.unlimited ? null : preview.rowsPerProject.remaining
+  const rowsCapByOperations = operationsRemaining === null ? null : Math.floor(operationsRemaining / algorithmCount)
+  const effectiveRowCap = [rowsRemaining, rowsCapByOperations].filter(n => n !== null).reduce(
+    (min, n) => min === null ? n : Math.min(min, n), null
+  )
+  const willTruncate = effectiveRowCap !== null && rowsToMatchCount > effectiveRowCap
+  const isPreviewQuotaExhausted = willTruncate && effectiveRowCap <= 0
```
paynejd: **An AI-only run is costed against the match-operations meter, and blocked by it.**

`algorithmCount` is derived from `algosSelected` unconditionally, but the dialog has an `algos` toggle (L54, "retrieve candidates") that the user can switch off to run AI analysis alone. With `algos === false` the run issues no `$match` calls and consumes no match operations — yet:

- `estimatedOperations` still reads `rows x algorithmCount`,
- `effectiveRowCap` is still capped by `rowsCapByOperations`,
- so with 0 operations left, `isPreviewQuotaExhausted` is `true` and **Submit is disabled** (L153-L158) for a run that would have cost nothing.

`MapProject.jsx:2091-2102` has the same bug on the execution side — it truncates `rowsToProcess` to 0 and runs nothing.

```suggestion
  const algorithmCount = algos ? Math.max(algosSelected.length, 1) : 0
  const estimatedOperations = rowsToMatchCount * algorithmCount
  const operationsRemaining = preview.matchOperations.unlimited ? null : preview.matchOperations.remaining
  const rowsRemaining = preview.rowsPerProject.unlimited ? null : preview.rowsPerProject.remaining
  const rowsCapByOperations = (operationsRemaining === null || algorithmCount === 0) ?
    null : Math.floor(operationsRemaining / algorithmCount)
  const effectiveRowCap = [rowsRemaining, rowsCapByOperations].filter(n => n !== null).reduce(
    (min, n) => min === null ? n : Math.min(min, n), null
  )
  const willTruncate = effectiveRowCap !== null && rowsToMatchCount > effectiveRowCap
  const isPreviewQuotaExhausted = willTruncate && effectiveRowCap <= 0
```

The `algorithmCount === 0` guard matters — without it the division yields `Infinity` and the cap silently disappears.

Two open questions this exposes, both of which need the oclapi2#900 answer:

- **Does an AI-only run still consume `mapper.rows_per_project`?** The suggestion above keeps `rowsRemaining` in the cap, which is the conservative reading. If AI analysis only meters `ai_assistant.calls`, the row cap shouldn't apply either.
- **`ai_assistant.calls` isn't checked here at all.** `getMapperPreview()` exposes it, but nothing in this dialog gates on it, so an AI-only run with an exhausted AI quota submits and fails server-side. Follow-up slice, or should it be in this one?

src/components/map-projects/AutoMatchDialog.jsx
```diff
@@ -143,6 +176,24 @@ const AutoMatchDialog = ({
...
+                  t('map_project.preview_estimate_no_rows_left') :
+                  willTruncate ?
+                  t('map_project.preview_estimate_will_truncate', {
+                    allowed: Math.max(effectiveRowCap, 0).toLocaleString(),
+                    requested: rowsToMatchCount.toLocaleString()
+                  }) :
+                  t('map_project.preview_estimate_note', {
+                    used: estimatedOperations.toLocaleString(),
+                    remaining: operationsRemaining !== null ? operationsRemaining.toLocaleString() : rowsRemaining.toLocaleString()
+                  })
```
paynejd, line 204: **Wrong unit in the estimate sentence when operations are unlimited.**

The fallback substitutes `rowsRemaining` into `preview_estimate_note`, whose text is:

> "This run will use {{used}} of your {{remaining}} remaining **match operations**."

When `operationsRemaining === null` (operations unlimited) but `rowsRemaining` is finite, the sentence reports a **row** count labelled as operations — and pairs it with `used: estimatedOperations`, which *is* an operation count. So the two halves of one sentence are in different units, and on a multi-algorithm run the "used" number can exceed the "remaining" number while nothing is actually wrong.

```suggestion
                  operationsRemaining !== null ?
                  t('map_project.preview_estimate_note', {
                    used: estimatedOperations.toLocaleString(),
                    remaining: operationsRemaining.toLocaleString()
                  }) :
                  t('map_project.preview_estimate_rows_note', {
                    used: rowsToMatchCount.toLocaleString(),
                    remaining: rowsRemaining.toLocaleString()
                  })
```

This needs a new key in all three locales alongside `preview_estimate_note` (en/es/zh `translations.json`) — suggested en copy:

```json
"preview_estimate_rows_note": "This run will use {{used}} of your {{remaining}} remaining preview rows.",
```

src/components/map-projects/MapProject.jsx
```diff
@@ -1527,12 +1543,18 @@ const MapProject = () => {
...
+          refreshMapperQuotaCache()
         if(response.data.url)
           history.push(response.data.url)
         if(!isAutoSave)
           baseSetAlert({severity: 'success', message: t('map_project.successfully_saved'), duration: 2000})
 
         APIService.new().overrideURL(response.data.url).appendToUrl('logs/').post({logs: {row_logs: rowLogsForSave, project_logs: savedProjectLogs}}).then(() => {})
+      } else if(status === 403 && errorData?.error_code) {
+        setPreviewLimit({errorCode: errorData.error_code, limit: errorData.limit, used: errorData.used})
+      } else {
```
paynejd: **This catch-all now toasts on throttling and on logout.**

Switching the save to `raw=true` means `handleAPIError` hands back the raw error instead of a pending request, so responses that were previously absorbed by `APIService` now fall through to this `else`:

- **429** — `handleAPIError` calls `notifyThrottlingListeners(error.response)` and returns `{handled: true, result: raw ? error : createPendingRequest()}` ([`APIService.js:72-74`](https://github.com/OpenConceptLab/oclmap/blob/657aa52/src/services/APIService.js#L72-L74)). The throttle UI already tells the user what happened; this adds a second, less informative error toast on top.
- **401** — the 401 branch triggers logout/redirect; an 8-second error toast rides along with the redirect.

Both are worse on **autosave**, where the user didn't initiate anything and gets an unexplained error toast mid-typing.

```suggestion
      } else if(status !== 429 && status !== 401) {
```

Separately: should `isAutoSave` suppress this toast entirely? An autosave failure is worth surfacing *somehow*, but an 8s error alert during editing is heavy — the `!isAutoSave` guard on the success path (L1550) suggests the intent was for autosave to stay quiet in both directions.

src/components/map-projects/PreviewLimitDialog.jsx
```diff
@@ -0,0 +1,65 @@
...
+// R12: at the limit, offer "request more access" (creates a lead) and "notify me
+// when plans launch" - there is no checkout until November, so neither button
+// takes a payment. One-time allowance (R2): no reset date is ever shown here.
+const CAP_COPY_KEY = {
+  mapper_rows_per_project_limit_reached: 'rows',
+  mapper_match_operations_limit_reached: 'match_operations',
+  mapper_projects_limit_reached: 'projects',
+  mapper_org_projects_denied: 'org_projects',
+  mapper_access_denied: 'access',
+}
```
paynejd, line 34: **This registry is missing codes the backends can actually emit, so they fall back to generic copy.**

oclapi2#900 defines seven mapper-facing codes. This map has five:

| `error_code` | source | mapped here |
|---|---|---|
| `mapper_projects_limit_reached` | `capabilities/constants.py:20` | ✅ |
| `mapper_rows_per_project_limit_reached` | `:21` | ✅ |
| `mapper_match_operations_limit_reached` | `:22` | ✅ |
| `mapper_access_denied` | `common/permissions.py:115` | ✅ |
| `mapper_org_projects_denied` | `:171` | ✅ |
| **`mapper_custom_algorithms_denied`** | `:144` | ❌ → `generic` |
| **`mapper_ai_assistant_denied`** | `:132` | ❌ → `generic` |
| **`capability_limit_reached`** (fallback) | `capabilities/views.py:59` | ❌ → `generic` |

`mapper_custom_algorithms_denied` is **reachable from this PR's own save path**: `CanUseCustomMapperAlgorithms` guards map-project create/update (`map_projects/views.py:35-39`), fires whenever `algorithms` contains a `type: 'custom'` entry, and a 403 from `onSave` lands at `MapProject.jsx:1554-1555` → `setPreviewLimit` → this dialog. A preview user who adds a custom algorithm and saves currently gets:

> **You've reached a preview limit**
> You've reached a preview limit for the Mapper.

instead of the backend's actual, much more useful message — *"Externally hosted algorithms are not available in preview."* No crash (the `generic` keys exist), just the wrong copy on a path that will be hit, since custom algorithms are exactly what preview users don't have.

```suggestion
const CAP_COPY_KEY = {
  mapper_rows_per_project_limit_reached: 'rows',
  mapper_match_operations_limit_reached: 'match_operations',
  mapper_projects_limit_reached: 'projects',
  mapper_org_projects_denied: 'org_projects',
  mapper_custom_algorithms_denied: 'custom_algorithms',
  mapper_ai_assistant_denied: 'ai_assistant',
  mapper_access_denied: 'access',
}
```

That needs four new keys per locale (en/es/zh), mirroring the backend's `denied_message` wording:

```json
"preview_limit_title_custom_algorithms": "Externally hosted algorithms aren't in the preview",
"preview_limit_body_custom_algorithms_no_count": "Externally hosted algorithms aren't part of the preview yet. Use OCL's built-in algorithms, or ask about more access below.",
"preview_limit_title_ai_assistant": "The AI Assistant isn't available yet",
"preview_limit_body_ai_assistant_no_count": "The AI Assistant isn't available for your account. Ask about more access below and we'll sort it out.",
```

Two related items, both tracked on the other PRs:

- **`mapper_ai_assistant_denied` is defined in oclapi2 but wired to no view**, and ocl-ai-assistant raises a bare `Http403()` with no `error_code` for that same denial — so today an unentitled AI user reaches `t('unknown_error')`, not this dialog. Raised on oclapi2#900 and ocl-ai-assistant#195; mapping it here is the third piece.
- **A not-entitled code is still needed.** Once #900 stops reporting `limit: null` as "limit reached", the new code wants an entry here too — otherwise the `_no_count` copy keeps telling unentitled users they spent an allowance they never had.

paynejd (commented):
## Cross-PR alignment pass — #60 / oclapi2#900 / ocl-ai-assistant#195

Follow-up to my review above, after reading all three `mapper-preview` PRs together. One correction to the dependency statement, and one new finding.

### Correction: the dependency is three-way, not two-way

**Deploy sequencing — all three must land together, in this order.** My original review here named only oclapi2#900; the full chain is:

1. **oclapi2#900** + the ocl_issues#2781 backfill — defines the capabilities, permissions, error codes and `/capabilities/consume/`.
2. **ocl-ai-assistant#195** — enforces `ai_assistant.calls` and reads `permissions`/`capabilities` from oclapi2. Against an oclapi2 that doesn't serve them it fails closed: total ai-assistant outage.
3. **oclmap#60** (this PR) — reads the meters and error codes both of the above produce.

Merging any of these independently produces a broken environment, not a partially-working one.

Concretely for this PR: **ocl-ai-assistant#195 is what makes `ai_assistant_calls_limit_reached` real.** `MapProject.jsx:4886` and `:4913` already branch on that code, but today the meter behind it never increments — #195's `CapabilitiesService.consume` raises `AttributeError` on every call (`self._headers` vs `_get_headers`) and a catch-all `except Exception` swallows it. So this PR's AI-quota handling is dead code until #195 lands with that fixed. Worth knowing when testing: an AI quota that never appears to exhaust is currently a **backend** bug, not a bug in this diff.

### New finding: the error-code registry is incomplete

`CAP_COPY_KEY` maps 5 codes; oclapi2#900 can emit more than that, and ai-assistant emits none at all. Detail inline on `PreviewLimitDialog.jsx`.

### Still open, and now confirmed as cross-repo decisions

- **Metering unit** — I've confirmed the direction: oclapi2 charges `units=len(rows)`, this PR estimates `rows x algorithmCount`. This PR matches #2762/#2782; the backend doesn't. Raised on #900 too.
- **`0 = unlimited` / `None = blocked`** — three implementations (`toMeter` here, `get_capability_limit` in oclapi2, `HasAIAssistantCapacity` in ai-assistant). My `autoMatchRows.js` comment above stands: this PR has `None` backwards in that one helper.
- **`limit: null` mis-messaging** — raised on #900, where the root cause is. It surfaces *here*, in `PreviewLimitDialog`'s `_no_count` copy: an unentitled user is shown "Your free preview's match-operation limit has been reached". Once #900 emits a distinct not-entitled code, this dialog needs the matching entry and copy.

paynejd (changes requested):
Formally marking **Request changes** — my two earlier reviews were submitted as comments, which understated where this stands. Nothing new here; this consolidates what's already inline so the PR state matches the findings.

**Blocking in this diff:**

1. **Bulk matching swallows a preview-limit 403 as a successful empty match** (`MapProject.jsx` ~L1954). `APIService` resolves error bodies rather than throwing, so rows are marked `algo_finished` with zero candidates, no alert, no dialog. This is the exact silent failure ocl_issues#2773 exists to fix; only the per-row path got the fix.
2. **A limit error on a plain row click leaves the row spinning forever** (`MapProject.jsx:3267`). `handlePreviewLimitError` returns before `markAlgo(-2)` and only restores the stage when a Refresh-button snapshot exists.
3. **The quota cache never refreshes on the ordinary row-click path** (`MapProject.jsx:3455` and three siblings). The refresh is gated on `clearRefreshRowStageSnapshot`, which only the Refresh button populates — so the meter goes stale on the most common path, which is also what makes (1) reachable.
4. **Existing sessions never see the feature** (`utils.js:275`). `refreshCurrentUserCapabilitiesCache` drops `permissions`, and every gate reads `permissions` — so `hasCapability` stays false until re-login. The endpoint already returns both fields; this is a one-line merge.

**Also needs resolving before merge** (detail inline): the AI-only run costed against the match meter and blocked by it; `getPreviewEligibleRowIndexes` treating a null limit as unlimited when everything else treats it as blocked; the wrong unit in `preview_estimate_note`; `CAP_COPY_KEY` missing codes the backend can emit; and the new catch-all toasting on 429/401.

**Merge order is not negotiable regardless of the above:** oclapi2#900 + the ocl_issues#2781 backfill → ocl-ai-assistant#195 → this PR. #900 has four open blockers of its own, and this PR reads the capabilities, meters and error codes it defines. Merging independently gives a broken environment, not a partial one.

Happy to re-review quickly once the four above are addressed — the architecture is right and `autoMatchRows.js` + the complete en/es/zh coverage are genuinely good. The open metering question (rows vs row-algorithm pairs) is a decision for ocl_issues#2762, not something to settle in this diff.

src/components/map-projects/MapProject.jsx
snyaggarwal: this was also the reason of AutomatchRun not reporting failed rows, added/tested that as well

src/common/utils.js
snyaggarwal: not falling back on `currentUser.permission` -- letting API dictate this.
Adding a check if the response has `capabilities` and/or `permissions` before updating it.

src/components/map-projects/AutoMatchDialog.jsx
snyaggarwal: added `ai_assistant.calls` in it

snyaggarwal: keeping rows_per_project for now -- matching api

src/components/map-projects/MapProject.jsx
snyaggarwal: adding `!isAutoSave`

src/common/utils.js
```diff
@@ -1122,7 +1139,39 @@ export const toCamelCase = str => {
...
+    const unlimited = limit === 0
+    return {limit, used, unlimited, remaining: unlimited ? null : Math.max((limit || 0) - used, 0)}
+  }
+  return {
+    hasAccess: hasCapability(user, 'users.mapper_use'),
+    hasAIAssistant: hasCapability(user, 'users.mapper_ai_assistant'),
+    hasCustomAlgorithms: hasCapability(user, 'users.mapper_custom_algorithms'),
+    hasOrgProjects: hasCapability(user, 'users.mapper_org_projects'),
+    projects: toMeter('mapper.projects'),
+    rowsPerProject: toMeter('mapper.rows_per_project'),
```
paynejd: **Rows are a cap, not a quota.** For this capability, `used` from `/user/` is a per-user lifetime counter: oclapi2 adds every completed row in any project and never subtracts. The server doesn't enforce against it; it checks the project's own rows at run creation. So `remaining` goes wrong as soon as a user deletes a project: on the new one, Auto Match says "No more rows are left in your preview quota" for good, while the server would accept the run.

The first N rows of the spreadsheet are the project. `getPreviewEligibleRowIndexes` already activates them, so what's left is:
- a notice at import when rows are dropped, e.g. "Preview projects hold 25 rows — we activated the first 25 of your 180"
- no more reads of `rowsPerProject.remaining`: `AutoMatchDialog.jsx` L90, the pre-truncation at `MapProject.jsx` L2114, and `MapperQuotaChip` (show the cap, not "rows left")

Re-running Auto Match then costs match quota only; OpenConceptLab/oclapi2#900 is changing the server side to match.

src/components/map-projects/AutoMatchDialog.jsx
```diff
@@ -45,25 +47,61 @@ const AutoMatchDialog = ({
...
+  // analysis is actually going to run.
+  const aiCallsRemaining = preview.aiAssistantCalls.unlimited ? null : preview.aiAssistantCalls.remaining
+  const rowsCapByAICalls = autoRunAIAnalysis ? aiCallsRemaining : null
+  const effectiveRowCap = [rowsRemaining, rowsCapByOperations, rowsCapByAICalls].filter(n => n !== null).reduce(
+    (min, n) => min === null ? n : Math.min(min, n), null
+  )
+  const willTruncate = effectiveRowCap !== null && rowsToMatchCount > effectiveRowCap
+  const isPreviewQuotaExhausted = willTruncate && effectiveRowCap <= 0
+  const isPreviewLimited = operationsRemaining !== null || rowsRemaining !== null ||
+    (autoRunAIAnalysis && aiCallsRemaining !== null)
```
paynejd: **The AI quota must not cap rows.** Folding `aiCallsRemaining` into `effectiveRowCap` makes the dialog promise a run that doesn't happen. With 6 AI calls left it says "We'll match the first 6 of the 25 rows you've selected", then the pre-truncation in `MapProject.jsx`, which ignores AI, matches all 25. With 0 left and AI ticked, it says "No more rows are left in your preview quota" and disables Submit, though matching would run fine without AI.

Drop AI from the cap, and keep how many rows get AI (`aiRowsCovered`) for the estimate line:

```suggestion
  // One $invoke per row. The AI quota never caps rows: once it runs out, the
  // rest of the run is matched without AI recommendations.
  const effectiveRowCap = [rowsRemaining, rowsCapByOperations].filter(n => n !== null).reduce(
    (min, n) => min === null ? n : Math.min(min, n), null
  )
  const willTruncate = effectiveRowCap !== null && rowsToMatchCount > effectiveRowCap
  const isPreviewQuotaExhausted = willTruncate && effectiveRowCap <= 0
  const rowsThisRun = willTruncate ? Math.max(effectiveRowCap, 0) : rowsToMatchCount
  const aiCallsRemaining = preview.aiAssistantCalls.unlimited ? null : preview.aiAssistantCalls.remaining
  const aiRowsCovered = (autoRunAIAnalysis && aiCallsRemaining !== null) ? Math.min(aiCallsRemaining, rowsThisRun) : null
  const isPreviewLimited = operationsRemaining !== null || rowsRemaining !== null || aiRowsCovered !== null
```

Preview AI is 6 `match-recommend` calls; once they're used, Auto Match still runs, just without AI.

src/components/map-projects/AutoMatchDialog.jsx
```diff
@@ -143,6 +182,29 @@ const AutoMatchDialog = ({
         <CloseIconButton onClick={onClose} />
       </DialogTitle>
       <DialogContent>
+        {
+          isPreviewLimited && rowsToMatchCount > 0 &&
```
paynejd: **This Alert can throw during render.** When rows and match operations are both unlimited, the AI quota is capped, "Run AI analysis" is ticked and the rows fit the AI calls left, none of the first three branches apply, and the last one (L201-204) calls `.toLocaleString()` on a null `rowsRemaining`. No launch group has that shape, but one per-user override does. The TypeError hands the page to the ErrorBoundary, and unsaved decisions are lost.

Render the Alert only when one of its branches applies:

```suggestion
          isPreviewLimited && rowsToMatchCount > 0 && (willTruncate || operationsRemaining !== null || rowsRemaining !== null) &&
```

The estimate spec and the rows-cap change replace this block, but it shouldn't ship as is.

src/components/map-projects/AutoMatchDialog.jsx
```diff
@@ -143,6 +182,29 @@ const AutoMatchDialog = ({
...
+                  willTruncate ?
+                  t('map_project.preview_estimate_will_truncate', {
+                    allowed: Math.max(effectiveRowCap, 0).toLocaleString(),
+                    requested: rowsToMatchCount.toLocaleString()
+                  }) :
+                  operationsRemaining !== null ?
+                  t('map_project.preview_estimate_note', {
+                    used: estimatedOperations.toLocaleString(),
+                    remaining: operationsRemaining.toLocaleString()
+                  }) :
```
paynejd: **Pre-run estimate.** Before a run starts, this dialog should say how far it gets:
- **Match operations:** active rows in scope × the algorithms that call `$match`, against the remaining match quota. Leave out scispacy, custom algorithms with their own `url`, and a bridge the user can't run (`canBridge` false); none of them spend match quota. Today `algorithmCount` (L87) and the pre-truncation at `MapProject.jsx` L2113-2116 divide by every configured algorithm, so a run with any of those selected is estimated high and cut short.
- **AI:** recommendations for the first `aiRowsCovered` rows, against the remaining AI quota.

For example: "25 rows × 2 algorithms = 50 of your 80 match operations. AI recommendations for the first 6 rows; the other 19 are matched without AI."

`MapProject.jsx` needs to pass in which algorithms will call `$match`, and the copy needs en/es/zh keys. People should see what a run costs before they spend it.

src/components/map-projects/MapProject.jsx
```diff
@@ -1911,6 +1962,17 @@ const MapProject = () => {
...
         );
+        // service.post() resolves (not throws) on a 403, so check explicitly.
+        // abortRef.current halts the rest of the run instead of draining it.
+        if(isPreviewLimitError(response)) {
+          forEach(rowBatch, __row => {
+            markAlgo(__row.__index, algo.id, -2)
+            log({action: 'algo_failed', extras: getAlgoLogExtras(algo)}, __row.__index)
+          })
+          handlePreviewLimitError(response)
+          abortRef.current = true
```
paynejd: **Running out mid-run isn't a Stop.** This sets the same `abortRef` as the Stop button, so a quota 403 is handled as a user cancel:
- the rerank pass bails (L2055-2058), and with it auto-propose on multi-algorithm runs
- `scheduleAutoSave('auto_match')` is skipped (L2185-2186), so the rows that did finish aren't saved
- `completeAutomatchRun` PATCHes the run as `cancelled`

Use a separate flag that stops sending `$match` but lets the run finish: rerank and auto-propose what finished, autosave, record the run as stopped for quota, and open the at-limit dialog once (OpenConceptLab/ocl_issues#2806). `completion_status` has no quota value yet, so PATCH `partial` and log the error code as the reason, or add a status in OpenConceptLab/oclapi2#900.

The user spent quota on those rows, so keep them.

src/components/map-projects/MapProject.jsx
```diff
@@ -3162,6 +3261,43 @@ const MapProject = () => {
...
+    if(Object.prototype.hasOwnProperty.call(refreshRowStageSnapshotRef.current, rowId)) {
+      delete refreshRowStageSnapshotRef.current[rowId]
+      return true
+    }
+    return false
+  }
+  const isPreviewLimitError = response =>
+    Boolean(response?.error_code && response.error_code.startsWith('mapper_'))
+
+  const handlePreviewLimitError = (response, rowId, algoId) => {
```
paynejd: **This never opens the at-limit dialog.** It toasts the backend's English `detail` and returns, so the limits preview users will actually hit never reach `PreviewLimitDialog`: `$match` from bulk runs, row clicks, Refresh and load more. A user who runs out gets a toast with no next step. It also doesn't stop a bulk bridge run: `fetchBulkBridgeCandidates` resolves and moves to the next row, so each remaining row sends another `$match` that 403s, plus a `GET /user/`.

OpenConceptLab/ocl_issues#2806 has the ACs: open the dialog once per run with the copy for that code, and stop sending requests that will also fail. The dialog's CTAs are `mailto:` links to me today; the ticket covers where they should go.

src/components/map-projects/MapProject.jsx
```diff
@@ -4769,10 +4935,20 @@ const MapProject = () => {
         setAnalysis(prev => ({...prev, [__index]: [...(prev[__index] || []), newEntry]}))
         return true
       } catch (err) {
+        if((err?.error_code || err?.response?.data?.error_code) === 'ai_assistant_calls_limit_reached') {
```
paynejd: **Stop AI on every AI denial, not only the limit.** OpenConceptLab/ocl-ai-assistant#195 answers `$invoke` with three 403 codes. Only `ai_assistant_calls_limit_reached` sets `aiQuotaExhaustedRef`. `ai_assistant_calls_not_entitled` and `mapper_ai_assistant_denied` fall through to the generic error, and `runBulkAIAnalysis` calls `$invoke` again for every remaining row. Treat all three the same way: stop AI, keep matching.

```suggestion
        if(['ai_assistant_calls_limit_reached', 'ai_assistant_calls_not_entitled', 'mapper_ai_assistant_denied'].includes(err?.error_code || err?.response?.data?.error_code)) {
```

This catch is the only path that sees these codes: `service.request` rejects on any non-2xx (`APIService.js` L169-172), so the `response?.error_code` check at L4911 never fires. OpenConceptLab/ocl_issues#2806 replaces the toast with per-code dialog copy.

src/components/map-projects/MapProjects.jsx
```diff
@@ -41,6 +41,12 @@ const MapProjects = () => {
   const history = useHistory()
 
   const user = getCurrentUser()
+  const [, setQuotaCacheVersion] = React.useState(0)
+  const refreshMapperQuotaCache = React.useCallback(() => {
+    refreshCurrentUserCapabilitiesCache(() => setQuotaCacheVersion(version => version + 1))
+  }, [])
+  const newProjectBlockReason = getNewProjectBlockReason(getMapperPreview())
```
paynejd: **Sessions from before the deploy can't start a project.** This gate (added in fe1f990) reads the cached user, and main never cached `permissions`: its `refreshCurrentUserCache` doesn't send `includeCapabilities`. Nothing refreshes the cache on this page (the left menu only does when opened), so every non-staff session from before the deploy sees New project disabled with "Mapper access is limited right now". Users with a project get out by opening it; a preview user with no project yet can't, short of logging in again. Refresh on mount:

```suggestion
  React.useEffect(() => { refreshMapperQuotaCache() }, [])
  const newProjectBlockReason = getNewProjectBlockReason(getMapperPreview())
```

Deploy day is exactly when every cached user is stale.

src/components/map-projects/MapProject.jsx
```diff
@@ -620,12 +633,22 @@ const MapProject = () => {
...
       createProjectFromTemplate()
     }
+    refreshMapperQuotaCache()
   }, [])
 
   React.useEffect(() => {
     setPermissionDenied(false)
   }, [params.projectId])
 
+  const newProjectBlockReason = (!params.projectId && !project?.id)
```
paynejd: **Same stale cache on `#/map-projects/new`.** The first render computes this from the pre-deploy cache, so it redirects to `/` and toasts "Mapper access is limited right now" before the mount refresh (L636) returns. Evaluate the gate once that refresh is back:

```suggestion
  const newProjectBlockReason = (!params.projectId && !project?.id && mapperQuotaCacheVersion > 0)
```

If the refresh fails, the gate stays open, and oclapi2 still enforces access and the projects cap on save.

paynejd (changes requested):
All four round-1 blockers are fixed.

**Blocking, before launch**

1. **Rebase on main.** `App.jsx` and `MapProject.jsx` conflict with the GA work from OpenConceptLab/ocl_online#137.
   - `App.jsx` imports: drop both `recordGAPageView` (main moved it into `GAService`) and `isInWaitlist` (this PR removed it).
   - `onCSVRowSelect`: main's `(csvRow, options = {})` signature plus your eligibility guard.
   - `onRefreshClick` is the one that matters. Keep main's GA line and your stage snapshot, and **drop** main's delete-this-row's-candidates block, which you removed on purpose in 94d5562. Keeping it still compiles and lints, so nothing will catch it.

   ```js
   const onRefreshClick = () => {
     GAService.recordActionEvent('MapProject', 'refresh_candidates')
     refreshRowStageSnapshotRef.current[rowIndex] = rowStageRef.current?.[rowIndex]
       ? {...rowStageRef.current[rowIndex]}
       : null
     fetchAllCandidatesForRow(getFirstAlgoDef()?.id, row, 0, undefined, undefined, undefined, true)
   }
   ```
   A test merge resolved this way passes ESLint, 199/199 unit tests, 6/6 integration tests and a production build.
2. **Rows are a cap, not a quota.** Activate the first N rows at import with a notice, and drop "rows left": it subtracts a per-user lifetime counter, so after a delete and recreate Auto Match says "No more rows" forever while the server would accept the run. Detail on `utils.js`.
3. **Pre-run estimate.** Before a run starts, the dialog should say how far it gets: match operations for the `$match` algorithms only, and which rows get AI. Spec on the estimate Alert.
4. **The AI quota never limits rows.** Folding it into the row cap makes the dialog promise "the first 6 of 25" and then match all 25, and disables Submit once AI is used up. The same block can crash the page. Suggestions on `AutoMatchDialog.jsx`.
5. **Running out mid-run isn't a Stop.** A quota 403 reuses the Stop flag, so finished rows aren't autosaved and the run is recorded `cancelled`. Inline.
6. **At-limit dialog.** It never opens for `$match`/`$invoke` limits, and its CTAs are mailto links to me → OpenConceptLab/ocl_issues#2806.
7. **Preview gating.** Org owners and external algorithms aren't hidden for preview users; `hasOrgProjects`/`hasCustomAlgorithms` are computed and never read → OpenConceptLab/ocl_issues#2805.
8. **No overspend.** ocl-ai-assistant will consume before invoking and dedupe by request key (OpenConceptLab/ocl-ai-assistant#195), so the idempotency key must be stable per (run, row). It includes the attempt number today ([`MapProject.jsx#L4885-L4889`](https://github.com/OpenConceptLab/oclmap/blob/eb7abeb2f3c8fa518e2754a4f7befbfde2b68609/src/components/map-projects/MapProject.jsx#L4885-L4889), outside the diff). `invokeTs` is set once per row before `retryWithBackoff`, so dropping `attempt` is enough:
   ```js
   'X-OCL-REQUEST-IDEMPOTENCY-KEY': `${params.projectId}-${__index}-${invokeTs}`,
   ```
   Also retry a 5xx from `$invoke` under the same key. [`isTransientNetworkError`](https://github.com/OpenConceptLab/oclmap/blob/eb7abeb2f3c8fa518e2754a4f7befbfde2b68609/src/services/APIService.js#L213-L217) only retries dropped connections, and a same-key retry is what makes a failed call free for the user. Stop the bulk AI loop on the two AI denial codes too, not only `_limit_reached` (suggestion inline).
9. **Sessions from before the deploy can't start a project.** fe1f990 gates New project on the cached user, which has no `permissions` for anyone who logged in before this ships, and nothing refreshes it on the list page or before the `/map-projects/new` redirect. Suggestions on `MapProjects.jsx` and `MapProject.jsx`.
10. **Don't persist whole HTTP responses in `candidates`.** `createAlgorithmResponse` stores the entire axios response object as `raw` ([`normalizers.js#L43`](https://github.com/OpenConceptLab/oclmap/blob/eb7abeb2f3c8fa518e2754a4f7befbfde2b68609/src/components/map-projects/normalizers.js#L43); callers pass `rawResponse: response`, e.g. [`MapProject.jsx#L3445`](https://github.com/OpenConceptLab/oclmap/blob/eb7abeb2f3c8fa518e2754a4f7befbfde2b68609/src/components/map-projects/MapProject.jsx#L3445)). `onSave` serializes it, request internals and all, and since 94d5562 every Refresh appends another one. Keep only what the UI reads (`data`, `status`), and drop the rest from existing state on load and before save. Details and the data cleanup are in OpenConceptLab/ocl_online#188.

**Decisions**
- A **cap** is how many of something can exist, freed on delete (projects, rows per project). A **quota** is operations per period (match operations, AI calls).
- Preview AI is 6 `match-recommend` calls. Once they're used, Auto Match still runs, without AI.
- Limits: `0` = unlimited, `-1` = blocked (the kill switch), no limit = not entitled. Show the last two as *not available*, never as a count: today the chip renders `0 of  …` for a missing limit and would render `-1`.
- Everyone is removed from `mapper-waitlist` at go-live, so nothing needs to key off it.

**Deploy order:** OpenConceptLab/oclinfrastructure#7 → OpenConceptLab/oclapi2#900 → Keycloak groups & membership → OpenConceptLab/ocl-ai-assistant#195 → OpenConceptLab/oclmap#60

**Follow-ups (non-blocking)**
- Every per-row action calls `GET /user/?includeCapabilities=true` (~22 oclapi2 queries) with no debounce ([`refreshMapperQuotaCache`](https://github.com/OpenConceptLab/oclmap/blob/eb7abeb2f3c8fa518e2754a4f7befbfde2b68609/src/components/map-projects/MapProject.jsx#L140-L143)).
- scispacy is gated on the `core_user` group name ([`MapProject.jsx#L528-L532`](https://github.com/OpenConceptLab/oclmap/blob/eb7abeb2f3c8fa518e2754a4f7befbfde2b68609/src/components/map-projects/MapProject.jsx#L528-L532)); make it a permission.
- `PreviewLimitDialog` uses raw MUI `Dialog` and `MapperQuotaChip` a plain `Tooltip`; use the design system's `common/Dialog` and `HTMLTooltip`, and flag anything they can't do as a design-system gap.

